PR #45, opened 2017-05-11
Update d2-ui commit revision to point to current dataset-configuration

### LINKED ISSUE #42: Review installation straight from a fresh clone repo

tokland:
Fixes #42


adrianq:
Easy... we just need to keep in mind we need to do this at least every time we release a new version

PR #46, opened 2017-05-17
Feature/wizard step4

tokland:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "d2-dataset-configuration",
-  "version": "0.2.2",
+  "version": "0.2.3",
```
adrianq: Let's go for 0.3.0. I am not gonna release this version but we can set the version now for the future release

src/DataSets/Forms/Sections.component.js
```diff
@@ -0,0 +1,382 @@
...
+import Translate from 'd2-ui/lib/i18n/Translate.mixin';
+import LinearProgress from 'material-ui/LinearProgress/LinearProgress';
+import FormHelpers from '../../forms/FormHelpers';
+import SearchBox from '../../SearchBox/SearchBox.component';
+import MultipleDataTable from '../../MultipleDataTable/MultipleDataTable.component';
+import LoadingStatus from '../../LoadingStatus/LoadingStatus.component';
+import ObserverRegistry from '../../utils/ObserverRegistry.mixin';
+import Action from 'd2-ui/lib/action/Action';
+import Sidebar from 'd2-ui/lib/sidebar/Sidebar.component';
+//import {Tabs, Tab} from 'material-ui-scrollable-tabs/Tabs';
```
tokland: The scrollable fork does not forward some styles so I could not customize the tabs. Should be temporal, we need custom styles and scrollable.

PR #47, opened 2017-06-07
Save dataset/sections + move model logic to DataSetStore

### LINKED ISSUE #18: Fourth step (Manage DataSet Sections)
- [x] Create Interface.
   - [x] Investigate current components in dhis graphical interfaces library to decide what would be the best strategy. This is somehow a tradeoff between UX and reusability
   - [x] It would be ideal to provide a single page interface. An accordion set of panel interface in which each panel corresponds to a section could be an option.
   - [x]  We would like to provide filters for dataelements/indicator per core competency and origin
   - [x] We would like to provide sorting per column

- [ ] Logic

   - [ ] Generate List of Sections based on Competencies from the first step. For each competency (data element group) we create two section: [Competency name] and  [Competency name] (B). 

   - [ ] [Competency name] section should contain all ‘output’ Data Element for that data element group. 
   - [ ] [Competency name] (B) section should contain all ‘outcome’ Indicator for that data element group (core competency). 
      - [ ] We will have to extract all dataelements from num and den. Just the first element of any UID.UID2.UID3.
      - [ ] For every indicator we will get the code and create a string like: [indicator_code] + "-C" . We will look for a data element with this code. Application should not fail if this data element can not be found.
Example:
Indicator:
GL-EB1A: % of enrolled learners who pass the final YEP program exam
Data Element:
GL-EB1A-C: comment: % of enrolled learners who pass the final YEP program exam
      - [ ] We will be adding as well the indicators to the dataset

   - [ ] Users should be allowed to select/unselect data elements/indicators per section. By default just ‘Global/Mandatory’ data elements/indicators are selected. DataElementOrigin defined the scope-> Global or Donor or Local/Country or... 

   - [ ] Users can also set Row Totals and Column Total per section

tokland:
Closes #18 
Also, I've moved model logic from FormSteps to DataSetStore.

### DIFF
--- a/src/DataSets/Forms/Save.component.js
+++ b/src/DataSets/Forms/Save.component.js
@@ -8,22 +8,22 @@ import CardText from 'material-ui/Card/CardText';
 const Save = React.createClass({
     mixins: [Translate],
 
-    states: {DATAENTRY: "DATAENTRY", SAVED: "SAVED", SAVE_ERROR: "SAVE_ERROR"},
-
     propTypes: {
         config: React.PropTypes.object,
-        data: React.PropTypes.object,
+        store: React.PropTypes.object,
         state: React.PropTypes.string,
         errors: React.PropTypes.arrayOf(React.PropTypes.string)
     },
 
     render() {
-        const {dataset} = this.props.data;
+        const {dataset} = this.props.store;
         const {state, errors} = this.props;
 
-        if (state == this.states.DATAENTRY) {
+        if (state === "DATAENTRY" || state === "SAVING") {
             return (
                 <div>
+                    {state === "SAVING" ? <LinearProgress /> : null}
+
                     <div>{this.getTranslation("wizard_presave_message")}</div>
 
                     <ul className="list-group">
@@ -39,7 +39,7 @@ const Save = React.createClass({
                     </ul>
                 </div>
             );
-        } else if (state == this.states.SAVE_ERROR) {
+        } else if (state === "SAVE_ERROR") {
             return (
                 <div className="alert alert-danger">
                     <div>{this.getTranslation("wizard_save_error_message")}</div>
@@ -49,7 +49,7 @@ const Save = React.createClass({
                     </ul>
                 </div>
             );
-        } else if (state == this.states.SAVED) {
+        } else if (state === "SAVED") {
             return (
                 <div className="alert alert-success">
                     <div>{this.getTranslation("dataset_saved")}</div>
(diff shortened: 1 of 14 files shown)


src/DataSets/Forms/Save.component.js
```diff
@@ -8,22 +8,22 @@ import CardText from 'material-ui/Card/CardText';
...
         errors: React.PropTypes.arrayOf(React.PropTypes.string)
     },
 
     render() {
-        const {dataset} = this.props.data;
+        const {dataset} = this.props.store;
         const {state, errors} = this.props;
 
-        if (state == this.states.DATAENTRY) {
+        if (state === "DATAENTRY" || state === "SAVING") {
```
adrianq: @tokland I saw you created a state dictionary in FormSteps component but it is not used here... it would be better if we use a unique dictionary for the states. I guess to pass it as a parameter can be a bit messy but maybe as mixin? What do you think?

src/models/DataSetStore.js
```diff
@@ -0,0 +1,146 @@
...
+            dataElementDecoration: true,
+            indicators: [],
+        });
+    }
+
+    getInitialState() {
+        const baseDataset = this.getInitialModel();
+        const baseAssociations = {
+            project: null,
+            coreCompetencies: [
```
adrianq: why this hardcoded?

adrianq: This means that we will have these two core competencies selected by default. Is this sth we discuss?

src/DataSets/Forms/Save.component.js
tokland: Good catch, I was a bit uneasy with this, but wasn't sure if it was worth to have a module/mixin. Probably a module with the constants is enough. Willdo.

src/models/DataSetStore.js
tokland: No, no, this was just for testing when we started at step 4, now it makes sense to remove it.

tokland:
> what do you think about adding a validation in the first step so that you need to select a core/competency at least? 

My bad, I hadn't tested this case with no sections. Not a technical issue, if we must disallow a dataset without sections, a validation here is the way to go. Otherwise, show a message on step 4 (instead of, well, blowing up :))

tokland:
Pushed. If you can confirm it makes no sense to select 0 core competencies, I'll add the validation.

adrianq:
Yes, it makes no sense. Please, add the validation...

tokland:
Done! Core competencies validated.

adrianq:
I have just merged the PR but will keep the issue open until we discuss tomorrow some of the problems you found...

PR #56, opened 2017-06-19
Check server on app start and timed intervals

### LINKED ISSUE #44: Manage non-logged app
Redirect to login or if not possible, show alert.

tokland:
Closes #44 

Checks on app bootup and every 10 seconds.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -35,6 +35,17 @@ render(
   </MuiThemeProvider>, 
   document.getElementById('app'));
 
+function safeGetUserSettings() {
+    const redirect = (err) => {
+        window.location.assign(config.loginUrl);
+        return Promise.reject(err || "Cannot connect to server");
+    };
+
+    return getUserSettings()
+        .then(settings => typeof(settings) === "object" ? settings : Promise.reject())
+        .catch(redirect);
+}
+
 function configI18n(userSettings) {
     const uiLocale = userSettings.keyUiLocale;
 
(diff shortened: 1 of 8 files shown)


tokland:
not_logged_in

src/app.js
```diff
@@ -35,6 +35,17 @@ render(
   </MuiThemeProvider>, 
   document.getElementById('app'));
 
+function safeGetUserSettings() {
+    const redirect = (err) => {
+        window.location.assign(config.loginUrl);
```
adrianq: @tokland we have this redirection to loginUrl here and in redirectToLogin method. The url path and the code that makes the redirection is redundant. Can we maybe extract this redirection method into an utils function? Then in the sessiondialog, redirectToLogin doesn't need to be a method hardcoded but a property that we pass to the component SessionDialog.  What do u think? Maybe I am being too picky, if so just let me know :)

adrianq:
@tokland just a minor comment...

Also for the record: we are finally checking the app every 15 mins and it is more a 'keep me alive' feature than an actual validation. We will have to adjust this number depending on the time out in the dhis2 instance of the client

adrianq:
@tokland Sorry to bother you again. I am merging this PR as  I am about to release a new version and I think it would be good to have this feature on it. I have created a [card](https://github.com/EyeSeeTea/dataset-configuration/issues/58) to discuss this minor refactor I mentioned in my review. If finally we decide to go for it we can implement it as part of the next sprint

PR #68, opened 2017-06-29
Fix/add missing field i18n translations

### LINKED ISSUE #66: Fix missing string
**public_can_view** missing string

tokland:
Closes #66 

Besides `public_can_view` I've added 4 more missing keys (obtained with an automatic search). There were some non-translated keys in `src/forms/form-fields/coordinate-field.js`; since it's not used I've decided just to remove the file.


tokland:
> I guess you have deleted coordinate-field because it was not used right?

Right, see first comment.

adrianq:
Yep, sorry, I am tired :)

Issue #64, opened 2017-06-28
Rename sections label


adrianq:
Rename sections. Example: 
Education => Education Outputs
Education (B) => Education Outcomes

PR #69, opened 2017-06-29
Feature/sections label and order

### LINKED ISSUE #63: Global Mandatory should be displayed at the top
On step 4, Global Mandatory should be displayed at the top

tokland:
Closes #63, #64 

DataElements sorted by (selected DESC, name ASC).

Nitpicking: The dataElements appear with the desired ordering on the first render. If the user clicks on a header to sort by that column, there is no way to return to the default sorting order other than changing of section (for that, we should add a mechanism to disable sorting on a column). Control+click, shift+click, something like that.  

dhis2-sections

### DIFF
--- a/src/models/Section.js
+++ b/src/models/Section.js
@@ -104,7 +104,7 @@ const getD2Section = (d2, section) => {
 
 const getOutputSection = (dataElements, opts) => {
     const {d2, config, coreCompetency} = opts;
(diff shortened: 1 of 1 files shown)


adrianq:
@adrianq to double check with the client if we would like to add a filter by selected data elements or sth like that

PR #70, opened 2017-06-29
Move redirectToLogin to dhis2 utils module

### LINKED ISSUE #58: Refactor loginurl redirection
From the [PR](https://github.com/EyeSeeTea/dataset-configuration/pull/56)

we have this redirection to loginUrl here and in redirectToLogin method. The url path and the code that makes the redirection is redundant. Can we maybe extract this redirection method into an utils function? Then in the sessiondialog, redirectToLogin doesn't need to be a method hardcoded but a property that we pass to the component SessionDialog. What do u think?

tokland:
Closes #58

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -25,6 +25,8 @@ injectTapEventPlugin();
 import routes from './router';
 import appTheme from './App/app.theme';
 import './App/App.scss';
+import {redirectToLogin} from './utils/Dhis2Helpers';
+
 
 // Render the a LoadingMask to show the user the app is in loading
 // The consecutive render after we did our setup will replace this loading mask
@@ -37,7 +39,7 @@ render(
 
 function safeGetUserSettings() {
     const redirect = (err) => {
-        window.location.assign(config.loginUrl);
+        redirectToLogin(config.siteUrl);
         return Promise.reject(err || "Cannot connect to server");
     };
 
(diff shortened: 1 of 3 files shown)


src/app.js
```diff
@@ -80,9 +82,10 @@ function startApp(d2) {
 // can use it to access the api, translations etc.
 getManifest('./manifest.webapp')
     .then(manifest => {
-        const baseUrl = process.env.NODE_ENV === 'production' ? manifest.getBaseUrl() : dhisDevConfig.baseUrl;
+        const baseUrl = process.env.NODE_ENV === 'production' ?
+            manifest.getBaseUrl() : dhisDevConfig.baseUrl;
+        config.siteUrl = baseUrl;
```
adrianq: siteUrl instead of baseUrl?

tokland: I was also perplexed.  It turns out `config.siteUrl` is used interally by d2 and must be set to `base/api/VERSION`. So I needed to create another key to point to the main site, thus the name.

adrianq: baseUrl is used internally you meant?

tokland: Internally, right. Note that this `config` comes from `import { init, config, getUserSettings, getManifest } from 'd2/lib/d2';`, it's a internal structure of d2.

adrianq: Ok, capito. You wrote siteUrl and this confused me. Merging...

tokland: Sorry, right, `baseUrl` is d2's, `siteUrl` is our new key.

Issue #19, opened 2017-03-23
Fifth step (Disaggregation)


adrianq:
We should reuse the current component and logic to override a category combo per dataset. How to access this feature can be found in this [section of the documentation](https://docs.dhis2.org/2.26/en/user/html/manage_data_set.html#override_dataelement_catcombo_in_dataset)

We should also implement the logic to create a data element category combination on the fly:
- [x] Application should be smart enough to decide if a new category combination needs to be created or we can reuse a previous one. App needs to hide options in which the category combination contains a category that is already part of the current one.
Dropdown list will display only category combination that can be appended to the default found. For instance: 
   - Valid use case
Current: DisplacementStatus/Sex
Option: Age
We create or reuse: DisplacementStatus/Sex/Age
   - Invalid use case:
Current: DisplacementStatus/Sex
Option: DisplacementStatus

- [x] We need to change the text in the dropdown list from: "override data ... " to "add disaggregate..."

adrianq:
Some doubts:
- [ ] What should be the behaviour if none category combination is set? And if there is an already set category combination but they user doesn't want to extra disaggregate? Is there a mandataroy category combination that we have to use anyway?

adrianq:
Updating card with today's discussion:
- No checkbox will be needed. Application should be smart enough to decide if a new category combination needs to be created or we can reuse a previous one. App needs to hide options in which the category combination contains a category that is already part of the current one.
Dropdown list will display only category combination that can be appended to the default found. For instance: 
   - Valid use case
Current: DisplacementStatus/Sex
Option: Age
We create or reuse: DisplacementStatus/Sex/Age
   - Invalid use case:
Current: DisplacementStatus/Sex
Option: DisplacementStatus

- We need to change the text in the dropdown list from: "override data ... " to "add disaggregate..."

PR #76, opened 2017-07-06
Feature/subsections and groups

### LINKED ISSUE #72: Implement subsections
- [x] Create Data Element Group Set and Indicator Group Set for Theme (when we use to call subsubsection)
- [x] Add Attribute for Subsubsection including indicator and elements. @adrianq to double check with the client what is the desired name
- [x] Implement javascript in custom/js app to reorganise sections and subsubsections
- [x] Modify dataset configuration layout to re-add subsections and subsubsections as columns

tokland:
Closes #72 

(#75 is not here, it will go to #19)

### DIFF
--- a/resources/customjscss/data-entry-contentscript.js
+++ b/resources/customjscss/data-entry-contentscript.js
@@ -12,140 +12,22 @@
         - Groups within a theme are grouped.
 */
 
+(function() {
+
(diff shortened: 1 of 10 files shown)


adrianq:
The code looks very good @tokland. Thank you! I am gonna merge it and do some more tests in development!

PR #78, opened 2017-07-10
Fix/dataelements with no override

### LINKED ISSUE #75: Set override category combination to "no override"

tokland:
Closes #75


src/models/DataSetStore.js
```diff
@@ -9,6 +22,7 @@ export default class DataSetStore {
         this.config = {
             categoryProjectsId: "MRwzyV0kXv9",
             categoryComboId: "GmXXE8fiCK5",
+            categoryComboDefaultId: "eCUWK7qTyeW",
```
adrianq: @tokland Why do we need this?

src/models/DataSetStore.js
```diff
@@ -126,12 +141,61 @@ export default class DataSetStore {
         this.updateFromAssociations(fieldPath, oldValue);
     }
 
+    saveDataset() {
+        return new Promise(async (complete, error) => {
+            // maintenance-app uses a custom function to save the dataset as it needs to do
+            // some processing  not allowed by dataset.save(). The code in this method is copied
+            // from maintenance-app/src/EditModel/objectActions.js
```
adrianq: @tokland and what is the difference from what we have before? There were some properties that were not being saved before?

src/models/DataSetStore.js
tokland: Indeed, it was used at some point in my tests,  but not anymore, removed: 793da63a

tokland:
8b48db43 sets renderAsTabs=TRUE

adrianq:
Sweet, ready to merge then?

src/models/DataSetStore.js
tokland: The most relevant part is how dataSetElements is saved, it need the dataset ID to be specified: `dataSet: { ...dataSet, id: dataSet.id || dataSetPayload.id },`. So it gets the UID from the server before saving, assigns the datSetElements (thus they are correctly set as "no override", and then the record is saved.

tokland:
Yup, ready to merge.

src/models/DataSetStore.js
adrianq: ahhhh... well, what a workaround anyway! Good that you catch it :)

adrianq:
Thanks @tokland!

PR #81, opened 2017-07-12
Feature/step 5 disaggregation

### LINKED ISSUE #19: Fifth step (Disaggregation)

tokland:
Closes #19

- DataSetElementCategoryComboSelectionDialog.component.js reused from maintenance-app (with our custom requirements).

- Prev/Next buttons on top and bottom (like the sections step)

dhis2-disaggregation


adrianq:
@tokland It looks good and seems to behave properly.

Issue #87, opened 2017-07-21
Random error on save


adrianq:
I was getting every now and then the following error. Couldn't identify when it was happening exactly.

![screenshot from 2017-07-21 01-24-29](https://user-images.githubusercontent.com/6850223/28444751-5b0cffee-6db7-11e7-877b-2cd292799483.png)

tokland:
There are two points where section.clone() is called, I've added a console.trace, check the JS console if this happen again, and hopefully you can also give the steps to reproduce it. I am surprised, the object is not undefined (it would say so in the trace), but has no clone(), which all d2 objects have.

I've aded the console.trace in the PR that fixes the grey-fields/sharing merge: 

https://github.com/EyeSeeTea/dataset-configuration/pull/90

adrianq:
@tokland I have merged everything and seems to be working fine. I have done around 10 tests and I only got an error on the first try. I haven't been able to reproduce it, even with the same parameters, so I guess it had to do with some cache issue. Anyway, I will keep an eye on it.

Once again, many thanks for this!

PR #90, opened 2017-07-21
Fix/fix merge greyed sharing

tokland:
Related: #87


adrianq:
@tokland Thanks for this and sorry for all the mess!

PR #101, opened 2017-07-28
Sharing has multi-select countries

### LINKED ISSUE #97: Modify sharing step
- [x] Multiple selection for sharing countries
- [x] Prefill logic:
    - If project, read country from project prefix code
    - If else organisation unit level 3 selected, add all OUs

Not to do by now: If non level 3 OU is selected, select associated countries from parents or children

tokland:
Fixes #97 

Related: EyeSeeTea/d2-ui/pull/8

(As previous PRs, change base to development before merging)


adrianq:
@tokland It seems to be working fine. I have created a couple of issues to tackle some bugs/enhancements. Some of them are related to this functionality but I don't think they are introduced in this PR:
https://github.com/EyeSeeTea/dataset-configuration/issues/111
https://github.com/EyeSeeTea/dataset-configuration/issues/110

I also created this one regarding non-level 3 org unit:
https://github.com/EyeSeeTea/dataset-configuration/issues/108
Not sure it is actually needed but I want to keep track of it. We can revisit it in the future.

Anyway, thanks for this! Merging it...

PR #104, opened 2017-08-02
Feature/23 settings panel

### LINKED ISSUE #23: Settings Panel
- [x] Create user rol for app admin
- [x] Add all configurable parameters as dropdown components
- [x] @adrianq to confirm we can use code instead of uid (yes, we can). But since we have a dropdown with names, we can still save uids.

tokland:
Fixes #23

We have the same limitation that we found in datadictionary, AFAIK we cannot use custom authorities (we can't, but they won't show up as selectable in the multi-select widgets). A "Dataset configuration admin" must have authority "M_dhis-web-maintenance-appmanager" or "ALL". Maybe there is a better way, we can talk about it when you return.

dhis2-setting


tokland:
Conflicts solved, ready to merge

src/app.js
```diff
@@ -71,9 +71,14 @@ function configI18n(userSettings) {
  * @param d2 Instance of the d2 library that is returned by the `init` function.
  */
 function startApp(d2) {
+    window.d2 = d2;
```
adrianq: @tokland why this?

adrianq:
@tokland This looks really great! Now that I see everything in place is really nice. Thanks!

I have aligned settings icon to the right. I think it looks better although it is true that is less visible.

Something minor (and probably a bit stupid and lazy on my side):
screenshot from 2017-08-14 22-21-48
why data element theme group set is not populated on init?

src/app.js
tokland: Debugging purposes so you can play with d2 on the js console.

tokland:
> why data element theme group set is not populated on init?

That's because this degs had not the default  expected code (GL_DETHEME_DEGROUPSET), it was created before all this, I've just set it. Default queries for settings:

https://github.com/EyeSeeTea/dataset-configuration/blob/development/src/models/Settings.js#L20

PR #106, opened 2017-08-10
Remove auto-bind library

tokland:
Breaks `yarn build`, removed.


adrianq:
Thanks for the quick fix!

PR #120, opened 2017-08-15
Set default sharing countries inconditionally on component mount

### LINKED ISSUE #111: If I choose a project, go to sharing, come back to step 1 and change project, sharings are not updated
Not sure what the behaviour should be....let's discuss on Monday

tokland:
Closes #111


adrianq:
@tokland Two things that are not working properly in my opinion:
- When I choose countries in step 3 I don't see them in the sharing step. They are not used to prefill sharing step anymore. If I am not wrong this used to work. 
- Something minor, if I choose a project and then I choose 'No value'. Prefilled values are still there. I think everything that is inferred from the project should be clear up, right?

tokland:
> When I choose countries in step 3 I don't see them in the sharing step. 

My bad, typo in the merge

>  I think everything that is inferred from the project should be clear up, right?

The policy when unseting a project is not removing associations (i.e. name). In this case, the orgunits assigned from the project remain and thus they appear in the sharing state. Do we better clear orgunits in this case?

adrianq:
I didn't even remember we decided that :) It is completely subjective but I would say we should clear up any association (name, input dates, org units, sharing...) What do you think? Do you remember why we decided to keep them? I am happy to merge this PR as it is and discuss this at our next meeting if you prefer....

tokland:
I don't really remember :) I don't have a strong opinion on that, let me clear the data and see how it goes.

adrianq:
Thanks for the quick fix!

PR #122, opened 2017-08-15
Move group-dataelements option to sections popover

### LINKED ISSUE #119: find a better place for 'group data elements in sections'
I guess the natural place would be in the sections tab (step 4) but this may need to reload this tab and I am not sure about how complex it would be...

tokland:
Closes #119


adrianq:
@tokland Really really nice!! I think it is super useful that just by clicking you can see the layout changing. It is fast and works nicely.

PR #158, opened 2017-09-15
Feature/140 user feedback

### LINKED ISSUE #140: User Feedback
Check platform or github issues (ws bridge?)

tokland:
Fixes #140 

- Using https://github.com/JoeyAndres/feedback-tool
- Currently, there is no way to use non-committed files at PRO env (config.js/config.json are only used in dev). So for now let's use a simple base64-encoded string for the token, it can be revoked at any time. What a malicious user could do: upload files to EyeSeeTeaBotTest/snapshots.
- When the DOM is complex (sections step, for example), the snapshot takes quite some time to complete (5-10 secs in my machine). 

screenshot from 2017-09-15 13-48-07
screenshot from 2017-09-15 13-48-18


adrianq:
@tokland the black out tool is not working, right?

adrianq:
Fixed. Changing mousedown to click seems to do the trick

PR #161, opened 2017-10-19
Sharing component with single/bulk objects

### LINKED ISSUE #116: Context Action (Bulk) -> Share
For Bulk actions, we will need a switch to change from Override to Append

tokland:
Closes #116 

- Base code from d2-ui/src/sharing (single edit for sharing) -> src/sharing
- On bulk edit, the form renders an intersection of all types of sharing:
  - external_access: `all?(external_accessses)`
  - public_access: `common_permission_for(external_accessses)`
  - user group accesses: `intersection_with_common_permissions(external_accessses for each group)`.
- The first render makes no changes yet, you have to change some checkbox (live saving).
- Bulk update strategies: switch with values _off_ (MERGE, default) and _on_ (REPLACE). Merge/replace strategy names follow the standard dhis2 terminology. Once you change the switch it applies (live saving) on the next change of checkbox you make (as always). I think we talked about having the switch on the listing component, but I think it makes more sense here, within the sharing component itself. 
- When only one element is selected in the form, the component works the same as before.

A note for debugging purposes, the URL to get the sharing of the object is (example):

`http://HOST:8080/api/26/sharing?id=Zfik2eM2Pa7&type=dataSet`

screenshot from 2017-10-19 10-40-58

screenshot from 2017-10-19 11-05-05


tokland:
First, merge EyeSeeTea/d2-ui/pull/9 into d2-ui/dataset-configuration and then update dataset-configuration's `yarn.lock`.

PR #177, opened 2017-11-07
Fix/allow repeated des in same section

### LINKED ISSUE #170: Data Element is added in more than one section
https://github.com/EyeSeeTea/dataset-configuration-blessed/issues/66

It looks like the same DE is used two times in an indicator. 

If this is the case we need to validate and just add it once.

If the problem is that the same DE is used in two sections we need to report to Troyel so he can fix the metadata

tokland:
Fixes #170 

(Depends of #175)


adrianq:
@tokland just out of curiosity, what was the problem here? were DEs used in multiple indicators?

PR #178, opened 2017-11-08
Add status column for data elements and indicators

### LINKED ISSUE #171: Add Status Column in Section 4
- [x] Add dataElement/indicator group set for status in settings
- [x] Hide Inactive
- [x] Sort first Active and then Phased Out
- [x] Highlight somehow phased out dataelement (darker background in the row, highlight 'phased out' label, etc)

tokland:
Closes #171


adrianq:
@tokland A few things I have noticed (I will update this list if I find something else):
- If I add GBV competency, I am getting the following error:
screenshot from 2017-11-14 13-35-38
I don't want to fix it myself because for Camp Management I can see an empty Outcome section so we are already managing empty outcome section. This should be a corner case I guess...
- In Education Outcome, I can see some 'Active Indicators (All)' at the top but a couple of them also at the bottom of the list. What is the reason for this?

adrianq:
Ok, I haven't found anything else. Just made some cosmetics changes...

adrianq:
@tokland I am merging this so I can test the last PR. If you can have a look at this and create a PR with these fixes it would be great. Thanks!

tokland:
Fixed in#182. The value was an object, now it's an array.

adrianq:
@tokland Thanks for the quick fix! Did you see my second comment? The first error is solved but I still can see the second one. Is this a bug or is there a reason for it?

adrianq:
Just created #185 so we don't forget about it. I am gonna release anyway today...

PR #191, opened 2017-11-21
Feature/expiry date setting

### LINKED ISSUE #188: "expiry days" refactor
- [x] Add default 'expiry days' to the admin panel
- [x] Set value by default to the one in the admin panel.
- [x] Add help to 'expiry days' explaining this field in step 2.

tokland:
Closes #188 

Depends on #190


src/Settings/Settings.component.js
```diff
@@ -95,36 +99,63 @@ const SettingsDialog = React.createClass({
...
                 label={this.getTranslation('save')}
+                disabled={!saveIsEnabled}
                 onTouchTap={this.save}
             /> : null,
         ];
 
+        const getTabCardProps = (section) => ({
+            fields: this.getFields(section),
+            onUpdateFormStatus: status => _.defer(this.onUpdateFormStatus, section, status),
+            onUpdateField: this.onUpdateField,
```
adrianq: @tokland Why are we adding these onUpdateFormStatus and onUpdateField? Out of curiosity...

tokland: Those two are props of d2-ui:FormBuilder. `onUpdateField` usage is clear, `onUpdateFormStatus` on the other hand is call wthen the validity of the form changes. That way you can enable/disable save/next buttons or similar.

PR #204, opened 2017-11-30
Use span instead of div for disaggregation de text to avoid ghost hovering

### LINKED ISSUE #200: Looks like there are some hidden labels when hovering in Dissagregation step (no permissions)
screenshot from 2017-11-29 22-14-42

tokland:
Fixes #200

### DIFF
--- a/src/forms/DataSetElementCategoryComboSelectionDialog.component.js
+++ b/src/forms/DataSetElementCategoryComboSelectionDialog.component.js
@@ -31,7 +31,7 @@ const enhance = compose(
 );
 
 const getCategoryOptions = (categoryCombo) => {
-    if (categoryCombo.isDefault) {
+    if (!categoryCombo || categoryCombo.isDefault) {
         return "";
     } else {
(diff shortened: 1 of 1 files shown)


src/forms/DataSetElementCategoryComboSelectionDialog.component.js
```diff
@@ -31,7 +31,7 @@ const enhance = compose(
 );
 
 const getCategoryOptions = (categoryCombo) => {
-    if (categoryCombo.isDefault) {
+    if (!categoryCombo || categoryCombo.isDefault) {
         return "";
     } else {
         return !categoryCombo ? null : collectionToArray(categoryCombo.categories)
```
adrianq: This will always be `collectionToArray(categoryCombo.categories)` right?

tokland: Indeed, now the second branch should be simplified. Will do on next task.

PR #206, opened 2017-12-06
On save error show also request json

### LINKED ISSUE #205: Show JSON request payload on save errors
Currently, it shows only the response, it will be easy to reproduce problems showing also the request

tokland:
Closes #205


tokland:
> are we still using validation_error and more_errors keys from the i18 files?

Good point, some merge must've gone wrong, #189 added this code and now it's gone away from development. I'll check it out.

adrianq:
As discussed, the fix to this last error (i18 missing) will go into #208. It looks like we screw it up in this merge https://github.com/EyeSeeTea/dataset-configuration/commit/ce3c355bc7c8c40d190bb25346867ae866941b4e

PR #231, opened 2018-01-05
Feature/show org units on save

tokland:
This branch was merged in development but somehow it didn't make its way to master (currently v0.7.1).


tokland:
Already in dev, will ship in next release, nothing to here.

PR #260, opened 2018-02-06
Show warning on sections step when many items are selected

### LINKED ISSUE #258: Show warning if too many items in sections
#de + #indicators > N show warning

tokland:
Closes #258 

screenshot from 2018-02-06 12-15-49


adrianq:
@tokland  I just increased a bit the warning threshold to 300. We can discuss what it is the right number tomorrow but this doesn't look too bad for me... not too slow at least

PR #295, opened 2018-03-23
Add help button to the wizard.

### LINKED ISSUE #273: Implement help button in each step

jordibc:
Closes #273
When adding a new dataset, it shows an icon with a question mark, that when clicked shows help related to the current step.


jordibc:
There are a couple of things that I think are missing:

* The actual help texts (and their translations), instead of the current placeholders in `src/DataSets/FormSteps.component.js`.
* The "help" button should not move the "cancel" button so mach to the left

jordibc:
Closes #273 .
help_button_273

src/Wizard/Wizard.component.js
```diff
@@ -5,14 +5,17 @@ import ToolbarGroup from 'material-ui/Toolbar/ToolbarGroup';
...
+import Dialog from 'material-ui/Dialog/Dialog';
+import FlatButton from 'material-ui/FlatButton/FlatButton';
 import Steps from 'react-steps';
 
 const Wizard = React.createClass({
     mixins: [Translate],
 
     propTypes: {
         fields: React.PropTypes.arrayOf(React.PropTypes.object),
-        buttons: React.PropTypes.arrayOf(React.PropTypes.object),
```
tokland: `this.props.buttons` is referenced in line 149, what's the idea of removing the propType?

src/Wizard/Wizard.component.js
```diff
@@ -61,10 +73,23 @@ const Wizard = React.createClass({
...
 
         return (
             <div>
+                <Dialog
+                    title={this.getTranslation('help')}
+                    actions={actions}
+                    open={this.state.helpOpen}
+                    onRequestClose={this._closeHelp}
+                >
+                {currentStep.help}
```
tokland: Indentation is missing in this line

src/Wizard/Wizard.component.js
```diff
@@ -98,7 +123,13 @@ const Wizard = React.createClass({
             .filter(({step}) => step.visible !== false);
     },
 
-    _renderButtons(step, showPrevious, showNext) {
+    _renderButtons(step, showPrevious, showNext, showHelp) {
+        const help = (
```
tokland: No need to change it here (the view is so trivial), but conditional views (which may be rendered or not) may go in a function so they are created only if necessary. Something like: `const help = () => ...`.

jordibc:
Thanks! I agree with all the comments. There is a new commit done addressing them... how should I proceed with this PR?

src/Wizard/Wizard.component.js
```diff
@@ -98,7 +124,15 @@ const Wizard = React.createClass({
             .filter(({step}) => step.visible !== false);
     },
 
-    _renderButtons(step, showPrevious, showNext) {
+    _renderButtons(step, showPrevious, showNext, showHelp) {
+        const renderHelp = () => (
+            <ToolbarGroup>
```
adrianq: Indentation is missing here...

src/DataSets/FormSteps.component.js
```diff
@@ -129,50 +129,58 @@ const DataSetFormSteps = React.createClass({
                 title: this.getTranslation("step_initial_configuration"),
                 component: InitialConfig,
                 props: props,
+                help: 'placeholder for initialConfig help',
```
adrianq: Can we add these strings to the multilanguage files?

adrianq:
Closes #273

src/Wizard/Wizard.component.js
jordibc: Sorry, I don't see where it's missing...

```
1. _renderButtons(step, showPrevious, showNext, showHelp) {
2.    const renderHelp = () => (
3.        <ToolbarGroup>
```

Line 2 starts with 4 extra spaces, and line 3 (the next level) has 4 more.

jordibc: Cool, I see it now, thanks!

adrianq:
@jordibc Thanks for this. It looks much better now. Two more things:
- These three terms (help_generalInformation,help_organisationUnit,help_landingPage) seem not to be following the convention right (underscore instead of camel case)
- When I try to get the help message for disaggregation and grey fields the interface gets frozen. Is it working for you?

jordibc:
@adrianq Yes, I looked at the ids for the terms, but I see now the titles and the translation files use underscored names. Fixed now.

As for the freezing, I hadn't noticed it before, but now I see it happens randomly at some step, as if there was a memory leak or some stalled computation -- it eventually resumes, but goes very slowly.

I couldn't find out why, I keep on looking.

jordibc:
@adrianq , @tokland found the problem of the freezing. The last commit should fix it.

PR #297, opened 2018-04-06
Feature/add logs 265

### LINKED ISSUE #265: Add logs mechanism
- [x] Logged Actions: create and any contextual action
- [x] Implement UI
- [x] Implement backend logic

jordibc:
Closes #265 

What this does is to add a log mechanism that registers the actions performed on one or several datasets, and a way to display the logs.

screenshot at 2018-04-06 14-52-49

There are two ways to view the logs:
* Click on the logs button (next to the settings-wheel button on the top right).
* Select "Logs" on the context action menu associated to one or several selected datasets.

The logs are saved on the `dataStore`, in a circular buffer.

Known Issues:
* Both editing a dataset or cloning it appear as "action: 'edit'" in the log. That is because they both use the same `Save` component, which is the one that logs the action. I'm not sure how to change it at this stage, or if it's worth it.
* The action of editing the organisation units is not being logged at the moment. That's because it uses a component from `d2-ui` (`OrgsUnitDialog`) that I don't know how to handle gracefully so it will log the proper result ("success" for a successful change of organisations, "failed" if there was a problem).


src/DataSets/DataSets.component.js
```diff
@@ -128,6 +132,25 @@ const DataSets = React.createClass({
         });
     },
 
+    getLogs(datasets) {
+        // Set this.state.logs to the logs that include any of the given datasets.
+        if (datasets === null) {
```
tokland: It's usually safer `if (!datasets)` (like in Python) or even `if (_(datasets).isEmpty())` if an empty list makes no sense as input.

src/DataSets/DataSets.component.js
```diff
@@ -128,6 +132,25 @@ const DataSets = React.createClass({
...
+            this.setState({logsObject: null});
+        } else {
+            this.context.d2.dataStore.get('dataset-configuration').then(store => {
+                return store.get('logs').catch(() => []);
+            }).then(logs => {
+                const idsSelected = new Set(datasets.map(ds => ds.id));
+                const hasIds = (log) => log.datasets.some(ds => idsSelected.has(ds.id));
+                const logsSelected = logs.filter(hasIds);
+                logsSelected.sort(dateSort);
```
tokland: Having lodash at hand, maybe: `const logsSelected = _(logs).filter(hasIds).sortBy("data").value();`

src/DataSets/DataSets.component.js
```diff
@@ -150,6 +173,16 @@ const DataSets = React.createClass({
         this.setState({settingsOpen: false});
     },
 
+    openLogs() {
+        this.context.d2.dataStore.get('dataset-configuration').then(store => {
```
tokland: We can move this logic to the `log.js` module so the implementation is hidden and can be easily reused.

src/DataSets/DataSets.component.js
```diff
@@ -344,6 +397,18 @@ const DataSets = React.createClass({
                                 onClose={listActions.hideDetailsBox}
                             />
                         : null}
+                    {
+                         this.state.logsObject ? (
+                             <Dialog
```
tokland: I don't want to slow down the release, but the amount of code involving logs probably deserve a separate component. You'd write here `<LogsDialog ...` and some methods and logic would end up there (getLogs, openLogs, ...). This would also avoid the `this.setState({logsObject: true})`, which is a bit of hack.

src/DataSets/context.actions.js
```diff
@@ -57,24 +60,35 @@ const canUpdate = (d2, datasets) => {
...
+    // Return a function that, when called with a dataset, logs the
+    // actionName and some info related to the dataset, and then calls
+    // f(dataset)
+    return (dataset) => {
+        //log(actionName, 'started', dataset);
+        // if uncommented, will log the fact that we tried to start the action.
+        f(dataset);
+    }
+}
+
```
tokland: But do we need logs on action starts? it's to keep the function as clean as possible, we try to minimize commented code.

src/DataSets/log.js
```diff
@@ -0,0 +1,53 @@
...
+
+    const d2 = await getD2();
+    const store = await d2.dataStore.get('dataset-configuration');
+
+    const logIndex = await store.get('logIndex').catch(() => 0);
+    const logs = await store.get('logs').catch(() => []);
+
+    const datasets = Array.isArray(dataset) ? dataset : [dataset];
+
+    const newLog = {date: Date(),
```
tokland: We favour this style for object literals: https://google.github.io/styleguide/jsguide.html#formatting-object-literals

src/DataSets/Forms/Save.component.js
```diff
@@ -34,11 +35,15 @@ const Save = React.createClass({
     },
 
     _redirectAfterSave() {
+        const {dataset, associations} = this.props.store;
+        log('edit', 'success', dataset);
```
tokland: You have access to the action being performed (add/edit/clone) in `this.props.store.action`.

(This solves your first item in "known issues")

src/DataSets/log.js
```diff
@@ -0,0 +1,53 @@
...
+}
+
+// Simple component to show a log entry.
+function LogEntry(props) {
+    return (<div key={props.date} style={{paddingBottom: '10px'}}>
+                <b>Date:</b> {props.date} <br />
+                <b>Action:</b> {props.action} <br />
+                <b>Status:</b> {props.status} <br />
+                <b>User:</b> {props.user.displayName} ({props.user.username})<br />
+                <b>Datasets:</b> {props.datasets.map(ds => `${ds.displayName} (${ds.id}) `)} <br />
```
tokland: This is rendering an array, we can add something like this: `.join(', ')`

src/DataSets/DataSets.component.js
jordibc: Thanks for the tip! The problem is, if I understand correctly, that it'd be
```
  ... .sortBy("date").value()
```
and because `date` is a string, it would not sort correctly.

The function dateSort already converts that string into an actual date so the comparison is correct.

I also have a small preference for vanilla javascript if it doesn't result in any significant performance drop or less readability (and happy to fix it if that's the case!).

src/DataSets/context.actions.js
jordibc: I agree, this came from a conversation between @adrianq and me, as we were not completely sure. I'd vote for removing this part.

src/DataSets/DataSets.component.js
tokland: Ah, yes, typo: `date`. 

You are correct, with the current format, it wouldn't work. This should work, though:  `sortBy(ds => new Date(ds.date))`

Usually dates are saved with ISO-8601 format, which is still a string, but the lexicographical comparison works. `const date = new Date().toISOString()`. However, this forces a parsing and reparsing to show a more human value, no need to do it here.

Lodash uses vanilla JS under the hood, there should be no performance hit.

src/DataSets/context.actions.js
adrianq: Then go ahead and just delete the comment! :)

jordibc: Great, done :)

PR #304, opened 2018-04-11
change logging messages

adrianq:


jordibc:
Looks all fine to me.

jordibc:
I'd suggest changing the format a bit though, to be uniform with the rest of the code:

`... action=='add'){` -> `... action === "add") {`

and so on, and add ";" at the end of the statements.

PR #312, opened 2018-04-17
Feature/autoselect project attribute 310

### LINKED ISSUE #310: Autoselect project attribute matching linked Data Set project
See equivalent blessed card for reference: https://github.com/EyeSeeTea/dataset-configuration-blessed/issues/98

tokland:
Closes #310 

The category.id/categoryOption.id relationships could be easily included in the custom form, but the JS code in a custom form is not executed until the form is rendered (and all selectors have value,) So the only way is to perform the same logic we apply in dataset-configuration: `if datasetName.includes(project.name) then select(project)`.

### DIFF
--- a/resources/customjscss/data-entry-contentscript.js
+++ b/resources/customjscss/data-entry-contentscript.js
@@ -8,8 +8,29 @@
         fixActionsBox();
     };
 
+    var selectProjectFromDataset = function(ev) {
+        var dataSetName = $("#selectedDataSetId option:selected").text();
+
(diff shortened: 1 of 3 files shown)


adrianq:
It is working fine after fighting for a while with the cache. Looking at the latest datasets they have created, it seems sometimes they decide to delete the bracket in the name. Anyway, for now, it is more than enough (it also works for the previous convention where 'DataSet' was not added to the name). Merging it...

PR #313, opened 2018-04-17
Feature/logs paging

jordibc:
This PR adds the capability to use paging for the logs.

At this moment, only the last two pages are retrieved, but the circular buffer for the log pages works, so the structure shouldn't change once we add the visualization of arbitrary pages to the Log Dialog.

I'd like to make this PR even before finishing the viewing part of the pages in the dialog so as to get your early feedback: do you think this paging approach makes sense? do you have a preference as to how to show the pages in the dialog?

### DIFF
--- a/src/DataSets/DataSets.component.js
+++ b/src/DataSets/DataSets.component.js
@@ -327,13 +327,6 @@ const DataSets = React.createClass({
             />,
         ];
 
-        const helpActions = [
-            <FlatButton
-                label={this.tr("close")}
-                onClick={this._closeHelp}
-            />,
-        ];
-
         const renderLogs = () => {
             const logs = this.state.logs;  // shortcut
             if (logs === null)
@@ -343,6 +336,13 @@ const DataSets = React.createClass({
             else
                 return logs.map(LogEntry);
         };
+
+        const helpActions = [
+            <FlatButton
+                label={this.tr("close")}
+                onClick={this._closeHelp}
+            />,
+        ];
 
         const renderHelp = () => (
             <div style={{float: 'right'}}>
--- a/src/DataSets/DataSets.component.js
+++ b/src/DataSets/DataSets.component.js
@@ -145,7 +145,7 @@ const DataSets = React.createClass({
             const title = this.tr("logs") + " (" + datasets.map(ds => ds.id).join(", ") + ")";
             this.setState({
                 logsObject: title,
-                logs: this.tr("logs_loading"),  // show while loading
+                logs: null,
             });
             getLogs().then(logs => {
                 const idsSelected = new Set(datasets.map(ds => ds.id));
@@ -184,7 +184,7 @@ const DataSets = React.createClass({
         const title = this.tr("logs") + " (" + this.tr("all") + ")";
         this.setState({
             logsObject: title,
-            logs: this.tr("logs_loading"),  // show while loading
+            logs: null,
         });
         getLogs().then(logs => {
             this.setState({logs: _(logs).orderBy('date', 'desc').value()});
@@ -334,6 +334,16 @@ const DataSets = React.createClass({
             />,
         ];
 
+        const renderLogs = () => {
+            const logs = this.state.logs;  // shortcut
+            if (logs === null)
+                return this.tr("logs_loading");
+            else if (_(logs).isEmpty())
+                return this.tr("logs_none");
+            else
+                return logs.map(LogEntry);
+        };
+
         const renderHelp = () => (
             <div style={{float: 'right'}}>
                 <IconButton tooltip={this.tr("help")} onClick={this._openHelp}>
@@ -423,10 +433,7 @@ const DataSets = React.createClass({
                                 onRequestClose={listActions.hideLogs}
                                 autoScrollBodyContent={true}
                             >
-                                {
-                                    !_(this.state.logs).isEmpty() ?
-                                        this.state.logs.map(LogEntry)
-                                    : this.tr("logs_none") }
+                                { renderLogs() }
                             </Dialog>)
                         : null }
                 </div>
--- a/src/DataSets/DataSets.component.js
+++ b/src/DataSets/DataSets.component.js
@@ -138,13 +138,17 @@ const DataSets = React.createClass({
         if (!datasets) {
             this.setState({logsObject: null});
         } else {
+            const title = this.getTranslation("logs") + " (" + datasets.map(ds => ds.id).join(", ") + ")";
+            this.setState({
+                logsObject: title,
+                logs: this.getTranslation("logs_loading"),  // show while loading
+            });
             getLogs().then(logs => {
                 const idsSelected = new Set(datasets.map(ds => ds.id));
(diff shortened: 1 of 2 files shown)


src/DataSets/log.js
```diff
@@ -1,44 +1,73 @@
...
+
+async function getLogs(pages, store) {
+    // Return the concatenated logs for the given pages (relative to
+    // logsPageCurrent) if they exist, or an empty list otherwise.
+    pages = pages || [-1, 0];
+    store = store ? store : await getStore();
+    const logsPageCurrent = await get('logs-page-current', 0, store);
+    const pagesNames = pages.map(n => 'logs-page-' + mod(logsPageCurrent + n, maxLogPages));
+    const logs = await Promise.all(pagesNames.map(name => get(name, [], store)));
+    return [].concat(...logs);
```
tokland: That's a flatting, right? lodash has `_.flatten(arrayofArrays)`

src/DataSets/log.js
```diff
@@ -1,44 +1,73 @@
...
+    store.set('logs-page-' + logsPageCurrent, logs);
+    if (logs.length >= maxLogsPerPage)
+        store.set('logs-page-current', mod(logsPageCurrent + 1, maxLogPages));
+}
 
-    store.set('logIndex', (logIndex + 1) % maxLogs);
-    store.set('logs', logs);
+async function get(key, defaultValue, store) {
+    // Return the value for key in the store.
+    return await store.get(key).catch(() => defaultValue);
```
tokland: Not a problem, but in simple functions like this, the `async`/`await` can be removed altogether.

src/DataSets/DataSets.component.js
```diff
@@ -74,6 +74,10 @@ const DataSets = React.createClass({
         };
     },
 
+    tr(text) {
```
tokland: Not used here, but if you ever need it, `getTranslation` has a second argument `variables = {}`, used as a namespace for string interpolation. In i18n files you then write: `key=my name is $$name$$`.

src/DataSets/DataSets.component.js
```diff
@@ -173,10 +181,21 @@ const DataSets = React.createClass({
...
         // Retrieve the logs and save them in this.state.logs, and set
         // this.state.logsObject to a description of their contents.
+        const title = this.tr("logs") + " (" + this.tr("all") + ")";
+        this.setState({
+            logsObject: title,
+            logs: null,
+        });
         getLogs().then(logs => {
-            this.setState({logs: _(logs).orderBy('date', 'desc').value().map(LogEntry)});
+            this.setState({logs: _(logs).orderBy('date', 'desc').value()});
```
tokland: Does it makes a local variable for the repeated expression?

src/DataSets/DataSets.component.js
```diff
@@ -173,10 +181,21 @@ const DataSets = React.createClass({
...
+        });
+    },
+
+    addLogPage() {
+        getLogs([-2]).then(logs => {
```
tokland: This `-2` is a bit cryptic, where does it come from? Is `addLogPage` used? I see no calls.

src/DataSets/log.js
```diff
@@ -1,44 +1,73 @@
...
-
-    const datasets = Array.isArray(dataset) ? dataset : [dataset];
+    const store = await getStore();
+    const logs = await getLogs([0], store);
+    logs.push(await logEntry(actionName, status, datasets));
+    setLogs(logs, store);
+}
 
-    const newLog = {
+async function logEntry(actionName, status, datasets) {
```
tokland: There are two functions called logEntry in this file

src/DataSets/log.js
```diff
@@ -1,44 +1,73 @@
...
-    const store = await d2.dataStore.get('dataset-configuration');
-
-    const logIndex = await store.get('logIndex').catch(() => 0);
-    const logs = await store.get('logs').catch(() => []);
-
-    const datasets = Array.isArray(dataset) ? dataset : [dataset];
+    const store = await getStore();
+    const logs = await getLogs([0], store);
+    logs.push(await logEntry(actionName, status, datasets));
+    setLogs(logs, store);
```
tokland: As a rule thumb, finish with a `return ...` so a promise is returned.

src/DataSets/log.js
jordibc: Yes, `logs` is an array with the logs from the requested pages (each one an array itself). I guess your solution is more readable, so I'm changing to it, thanks!

src/DataSets/log.js
jordibc: You are right, it's unneeded and I was just being a bit paranoid. Removed (the simpler the better), thanks!

src/DataSets/DataSets.component.js
jordibc: Neat, I didn't know. I'm adding it to the function signature.

I'm happy you didn't think it was too silly to add this tr() function, btw. :)

src/DataSets/DataSets.component.js
jordibc: I'm not sure I follow your comment.

If you mean the `const title = ...`, I added to emphasize that the object just contains the title.

If you mean why I removed the `.map(LogEntry)` part, it's so we store the logs themselves and can concatenate (and filter and sort) more when more pages are requested.

src/DataSets/DataSets.component.js
jordibc: Yes, sorry, this is really a placeholder for the part that will load more log pages when requested. The `-2` at this moment makes reference to the page previous to the already loaded ones (by default, pages `[-1, 0]` are loaded).

I should at the very least added a comment, if not removed the function completely.

src/DataSets/log.js
jordibc: Yes, one `logEntry` (creates a new log entry object) and a function-component `LogEntry` (capital `L`), that displays a log entry.

I'm changing the first one to `makeEntry` to avoid confusion.

src/DataSets/log.js
jordibc: I don't follow this one quite well...

Do you mean to `return setLogs(logs, store);` (and make `setLogs` end with `return store.set(...);` too)?

jordibc:
@tokland Thanks!

As for the sharing dialog creating a log even without doing any changes, you are right, I will investigate how to fix it.

src/DataSets/DataSets.component.js
tokland: Dismiss my comment, I was tired, I thought they were two different lines!

src/DataSets/log.js
tokland: Perfect!

src/DataSets/log.js
tokland: Exactly. When async functions return promises, callers can thus decide how to chain them or do error control (then/catch).

jordibc: Cool, sounds great. I've just added it (the last one being a `Promise.all()` because two things need to be updated, the page itself and the page index).

PR #318, opened 2018-04-24
Feature/logs paging active

jordibc:
Add a "get more logs" button that loads logs from previous pages.

There are no changes to the dataStore configuration, we just use it appropriately in the main application to show older logs on request.


src/DataSets/log.js
```diff
@@ -1,55 +1,87 @@
...
+    // It returns null when there are no more log pages.
+    if (pages.every(n => n >= maxLogPages))
+        return null;
 
-    if (logs.length < maxLogs)
-        logs.push(newLog)
+    const store = await getStore();
+    const logsPageCurrent = await store.get('logs-page-current').catch(() => 0);
+    const pagesNames = pages.map(n => 'logs-page-' + mod(logsPageCurrent - n, maxLogPages));
+    const logs = await Promise.all(pagesNames.map(name => store.get(name).catch(() => null)));
```
tokland: I've seen there is a way to avoid 404 on non-existing pages. The store has a method to get the current keys. Thus we could also remove the `catch`, which might mask real problems (network issues/implementation bugs). Just an idea:

```
const storeKeys = new Set(await store.getKeys());
const existingPageNames = pagesNames.filter(name => storeKeys.has(name));
const logs = await Promise.all(existingPageNames.map(name => store.get(name)));
```

src/DataSets/DataSets.component.js
```diff
@@ -303,21 +332,37 @@ const DataSets = React.createClass({
 
         const logActions = [
             <FlatButton
-                label={this.getTranslation('close')}
+                label={(this.state.logsPageLast < 0 ? this.tr("logs_no_older") : this.tr("logs_older")) +
```
tokland: See later comment about how to avoid 404 for non-existing pages. I was thinking we could use the same idea: since we already know the existing pages in the store, we can hide the button if we are sure there is no more data to load.

src/DataSets/DataSets.component.js
```diff
@@ -303,21 +332,37 @@ const DataSets = React.createClass({
 
         const logActions = [
             <FlatButton
```
tokland: A question about the UX, @adrianq : does it make sense to move this action inside the logs box, as center button? this way the user only sees it when it really makes sense.

adrianq:
@tokland I talked to @jordibc and it looks like he is not gonna have time before Tuesday for this. Can you please implement the changes yourself in this branch? I will make a final review afterwards and have a look at the UX...

tokland:
I've applied all my changes.

Also implements the "sharing: save only on changes" feature, which requires https://github.com/EyeSeeTea/d2-ui/pull/20

tokland:
Task "Sharing: Move logging to parent component, so it logs only once when dialog is closed." -> Done. Ready to test again.

PR #337, opened 2018-05-25
Feature/pagination 332

### LINKED ISSUE #332: Fix pagination
- [x] Fix pagination
- [x] Show human date in column lastUpdated (like logs: 23/05/2018, 13:54:03)

tokland:
Closes #332 

screenshot from 2018-05-25 11-16-16

### DIFF
--- a/src/DataSets/DataSets.component.js
+++ b/src/DataSets/DataSets.component.js
@@ -84,6 +84,7 @@ const DataSets = React.createClass({
         return {
             config: null,
             isLoading: true,
+            page: 1,
             pager: { total: 0 },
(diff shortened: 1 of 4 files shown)


adrianq:
@tokland Something very minor but while testing this I just noticed something:
screenshot from 2018-05-25 12-25-36
screenshot from 2018-05-25 12-25-19

Could it be that the Last Updated format is different depending if the dataset was created with DC or the vanilla app?

tokland:
That's really weird, some strings the date constructor cannot parse. I've refactored it to use separate fields, now it should always work.

PR #338, opened 2018-05-25
Feature/disaggregation multiselect

### LINKED ISSUE #333: Multiselect on disaggreation dropdown
- [x] User can select many categories for a catcombo
- [x] Make sure catCombos are reused (in any order of categories).

tokland:
Closes #333


adrianq:
It looks good to me. I have deeply tested with existing catcombos, new ones, editing, using default, single and multiple cat combos, etc and seems to be working fine. The only very minor thing is whether we should use a slash instead of a comma when multiple categories are added. Anyway, I am merging this and we can discuss it tomorrow

PR #371, opened 2019-04-12
Add async validation for dataset name uniqueness

### LINKED ISSUE #369: Do not allow duplicated names
- [x] Don't allow duplicated names in form
- [x] Check also if there are any duplicated names in the DB at the moment

tokland:
= PR. Pull request template

### :pushpin: References

* **Issue:** Closes #369 

### :memo: Implementation

- FormBuilder bites again: we can force sync validations (like we do on render), but not async validations. Workaround: I added the async validations on the field (will be executed on focus out of the field), and also run the validations explicitly when the Next button is pressed. In this last case, the errors are shown above the component (like we do in other steps).

### :art: Screenshots

Screenshot from 2019-04-12 19-10-41

### DIFF
--- a/src/DataSets/Forms/GeneralInformation.component.js
+++ b/src/DataSets/Forms/GeneralInformation.component.js
@@ -63,6 +63,9 @@ const GeneralInformation = React.createClass({
         if (existsDataSetWithName) {
             throw this.getTranslation("dataset_name_exists");
         }
+        else{
+            this.setState({ error: null });
+        }
     },
 
     _renderForm() {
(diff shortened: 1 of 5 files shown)


adrianq:
@tokland all good, just made a simple change https://github.com/EyeSeeTea/dataset-configuration/pull/371/commits/f9cab840e280216ea3fc24f78e173f7b110a27ba to avoid an error in a corner case: introduce duplicated name > press next > error shows up on top of the page > change name for a non-duplicated name > error doesn't disappear from the top of the page.

PR #376, opened 2019-04-25
Fix uniqueness on dataset edit

tokland:
This PR fixes a bug with the dataset.name uniqueness on edit mode.


tokland:
Duplicated, closing.

PR #391, opened 2019-05-30
[Disaggregation] Group non-comment data elements by indicator

### LINKED ISSUE #381: Disaggregation Consistency
Indicators are based on data elements (numerator and denominator). Force consistency of disaggregation for numerator and denominator data elements for any given indicator. For example, it should not allow any user to choose default in numerator and sex (male/female) in denominator.

tokland:
Closes #381 

Implementation:

- Group dataElements from the the same indicator, exclude dataElement if its code ends with `-C` (comment). A change of disaggregation affects both.
- On search: Show full group (numerator and denominator, when grouped).

Notes:

On long data element names, we may have a line break because of the length and then another for the second dataelement. Maybe we can add a visual cue.

image

Filter:

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "d2-dataset-configuration",
-    "version": "1.1.0",
+    "version": "1.1.1",
     "description": "Dataset Configuration User Interface",
     "main": "src/index.html",
     "scripts": {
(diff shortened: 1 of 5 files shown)


adrianq:
@tokland  PR seems to be working fine but I just noticed a bug we introduced in our last release. There was an exception when the user doesn't introduce nor the start date neither the end date for the dataset. I think I fixed it here https://github.com/EyeSeeTea/dataset-configuration/pull/391/commits/5f0e30778868b6772758eb779b6dc3ba2a8d4905 but it would be great if you can also test it a bit more before merging. In particular, strange cases as for instance when no dates are introduced, only one of them and the same with the output and outcome dates... I will wait for your confirmation before merging it!

tokland:
Indeed, there way missing this check. Another option is to default the object `const obj = (periodDates[type] || {})[periodYear] || {}`, but it's ok as it is.

PR #469, opened 2024-10-17
Associate projects to data sets

eperedo:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/86960mp8j

### :memo: Implementation

- [x] Add tabs to show list for dataSets and projects
- [x] Script to migrate project information from category options to custom attribute (metadata attached in issue)

[Project-Configuration-app_tabs_datasets_projects.webm](https://github.com/user-attachments/assets/30be6306-3f02-4c6f-a185-a514df158073)

Run script:

```bash
yarn run migrate-ds-project -u 'http://localhost:8080' --username 'username' --password 'mypassword'
```

#86960mp8j


package.json
```diff
@@ -28,6 +28,8 @@
         "d2-manifest": "1.0.0",
         "eslint-plugin-no-relative-import-paths": "^1.5.3",
         "font-awesome": "4.7.0",
+        "lodash.keyby": "4.6.0",
+        "lodash.mapvalues": "4.6.0",
```
tokland, line 32: est-lodash cannot do that?

src/domain/repositories/MetadataRepository.ts
```diff
@@ -0,0 +1,6 @@
+import { MetadataItem } from "$/domain/entities/MetadataItem";
+import { FutureData } from "$/data/api-futures";
```
tokland: Move FutureData to the domain, we cannot reference the data layer from here.

src/domain/entities/MetadataItem.ts
```diff
@@ -0,0 +1,8 @@
+import { NamedRef } from "$/domain/entities/Ref";
+
+export type NamedCodeRef = NamedRef & { code: string };
+
+export type MetadataItem = {
```
tokland: That names is too generic, from the name we cannot imagine what it contains

src/domain/entities/MetadataItem.ts
```diff
@@ -0,0 +1,8 @@
...
+    attributes: { project: NamedCodeRef; createdByApp: NamedCodeRef };
```
tokland: I'd say "attributes" is a implementation detail, so it should appear referenced in the domain layer.

src/data/repositories/ProjectD2Repository.ts
```diff
@@ -0,0 +1,110 @@
...
+import { MetadataItem } from "$/domain/entities/MetadataItem";
+import { Future } from "$/domain/entities/generic/Future";
+
+export class ProjectD2Repository implements ProjectRepository {
+    private d2DataSetApi: DataSetD2Api;
+    constructor(private api: D2Api, private metadata: MetadataItem) {
+        this.d2DataSetApi = new DataSetD2Api(this.api, metadata);
+    }
+
+    getAll(): FutureData<Project[]> {
```
tokland, line 24: As a general rule, be careful with getAll methods that return an array, it's unbounded. There are solutions with Generators, but it's complex. Not saying it has to be changed here, but so we are aware of this.

src/domain/usecases/GetProjectsUseCase.ts
```diff
@@ -0,0 +1,12 @@
+import { FutureData } from "$/data/api-futures";
+import { Paginated } from "$/domain/entities/Paginated";
+import { Project } from "$/domain/entities/Project";
+import { GetDataSetOptions } from "$/domain/repositories/DataSetRepository";
+import { ProjectRepository } from "$/domain/repositories/ProjectRepository";
+
+export class GetProjectsUseCase {
+    constructor(private projectRepository: ProjectRepository) {}
+    execute(options: GetDataSetOptions): FutureData<Paginated<Project>> {
```
tokland, line 10: blank line

tokland (commented):
I've tried to review only new things, in the future I should be more careful in the order of review so we don't have PRs with common code, which makes review harder.

src/data/repositories/D2ApiCategoryOption.ts
```diff
@@ -1,27 +1,23 @@
...
+                    filter: { id: { in: categoryOptionsIds } },
+                    fields: { id: true, displayName: true, lastUpdated: true },
+                    paging: false,
                 })
-                .value()
-        ).map(d2Request => {
+            ).map(response => [response]);
+        }).map(d2Request => {
             return d2Request.flatMap(d2Response => d2Response.objects);
         });
```
tokland, line 20: Is this equivalent?

```typescript
        return chunkRequest(ids, categoryOptionsIds => {
            return apiToFuture(
                this.api.models.categoryOptions.get({
                    filter: { id: { in: categoryOptionsIds } },
                    fields: { id: true, displayName: true, lastUpdated: true },
                    paging: false,
                })
            ).map(response => response.objects);
        });
```

src/data/repositories/D2ApiConfig.ts
```diff
@@ -0,0 +1,90 @@
...
+            });
+
+            const dataElementGroupSets = d2Response.dataElementGroupSets.map(
+                (d2Attribute): D2NamedCodeRef => {
+                    return { id: d2Attribute.id, name: d2Attribute.name, code: d2Attribute.code };
+                }
+            );
+
+            return {
+                attributes: { ...this.buildAttributes(attributes) },
```
tokland: No need to pack/unpack

src/data/repositories/D2ApiConfig.ts
```diff
@@ -0,0 +1,90 @@
...
+
+            const categories = d2Response.categories.map((d2Attribute): D2NamedCodeRef => {
+                return { id: d2Attribute.id, name: d2Attribute.name, code: d2Attribute.code };
+            });
+
+            const dataElementGroupSets = d2Response.dataElementGroupSets.map(
+                (d2Attribute): D2NamedCodeRef => {
+                    return { id: d2Attribute.id, name: d2Attribute.name, code: d2Attribute.code };
+                }
+            );
```
tokland: The 3 blocks are not changing anything, right?

src/data/repositories/D2ApiConfig.ts
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    private buildAttributes(attributes: D2NamedCodeRef[]): D2Config["attributes"] {
+        return {
+            project: this.getOrThrow(attributes, metadataCodes.attributes.project),
+            createdByApp: this.getOrThrow(attributes, metadataCodes.attributes.createdByApp),
+        };
+    }
+
+    private getOrThrow(modelData: D2NamedCodeRef[], code: string): D2NamedCodeRef {
```
tokland: [subjective] Generic functions that do not use anything from the instance can go to the root level.

src/data/repositories/DataSetD2Repository.ts
```diff
@@ -48,137 +50,149 @@ export class DataSetD2Repository implements DataSetRepository {
...
+                                displayName: true,
+                                path: true,
+                            },
+                        },
+                    })
+                ).map(d2Response => [d2Response]);
+            });
+
+            return $requests.map(d2Response => {
+                const allDataSets = d2Response.flatMap(d2Response => d2Response.objects);
```
tokland: same

src/data/repositories/DataSetD2Repository.ts
```diff
@@ -48,137 +50,149 @@ export class DataSetD2Repository implements DataSetRepository {
...
+                            throw Error(`Cannot find dataSet: ${dataSetId}`);
+                        }
+
+                        const existingAttributes = existingDataSet?.attributeValues;
+
+                        const result = {
+                            ...(existingDataSet || {}),
+                            ...this.buildD2DataSet(dataSet, existingAttributes, config.attributes),
+                        };
+                        delete result.sharing;
```
tokland: [FP, immutability] Not that it matters as it's an internal object, but we try not to mutable. If you want to change it: Rec.omit or unpack.

src/data/repositories/DataSetD2Repository.ts
```diff
@@ -48,137 +50,149 @@ export class DataSetD2Repository implements DataSetRepository {
...
+                    { dataSets: dataSetIds.map(id => ({ id })) },
+                    { importStrategy: "DELETE" }
+                )
+            ).map(response => {
+                const allErrors = response.typeReports.flatMap(typeReport =>
+                    typeReport.objectReports.flatMap(objectReport =>
+                        objectReport.errorReports.flatMap(errorReport => errorReport.message)
+                    )
+                );
+                if (allErrors.length > 0) throw new Error(allErrors.join("\n"));
```
tokland: No throw in a Future block, return Future.error instead (in a flatMap)

src/domain/entities/Permission.ts
```diff
@@ -1,7 +1,15 @@
-export type Permission = {
-    canRead: boolean;
-    canWrite: boolean;
-    noAccess: boolean;
-};
+import { Struct } from "$/domain/entities/generic/Struct";
+
+export type PermissionAttrs = { read: boolean; write: boolean };
 
 export const NO_ACCESS_NOTATION = "--------";
```
tokland: Move somewhere else, as now the domain knows nothing about octal format.

src/domain/entities/Permission.ts
```diff
@@ -1,7 +1,15 @@
...
+
+export class Permission extends Struct<PermissionAttrs>() {
+    noAccess(): boolean {
+        return this.read === false && this.write === false;
```
tokland: let's use the more standard `!someBoolean`

src/domain/entities/Sharing.ts
```diff
@@ -1,8 +1,8 @@
-import { Id, OctalNotationPermission } from "$/domain/entities/Ref";
+import { Id } from "$/domain/entities/Ref";
 import { Struct } from "$/domain/entities/generic/Struct";
 
 export type SharingAttrs = {
-    publicAccess: OctalNotationPermission;
+    publicAccess: string;
```
tokland: publicAccess as a string? shouldn't be a boolean or similar?

src/webapp/components/dataset-table/DataSetTable.tsx
```diff
@@ -223,60 +42,58 @@ export const DataSetTable: React.FC = React.memo(() => {
...
+    }, []);
 
     const goToCreateDataSet = React.useCallback(() => {
-        history.push("/dataSets/create");
-    }, [history]);
+        navigateTo("createDataSets");
+    }, [navigateTo]);
+
+    const renderActions = () => {
+        switch (action) {
```
tokland, line 71: nice

src/webapp/components/dataset-wizard/utils.tsx
```diff
@@ -0,0 +1,32 @@
+import i18n from "$/utils/i18n";
+import { IndicatorsDataSet } from "$/webapp/components/dataset-wizard/IndicatorsDataSet";
+import { SetupDataSet } from "$/webapp/components/dataset-wizard/SetupDataSet";
+import { ShareOptionsDataSet } from "$/webapp/components/dataset-wizard/ShareOptionsDataSet";
+import { SummaryDataSet } from "$/webapp/components/dataset-wizard/SummaryDataSet";
+
+export function useDataSetSteps() {
+    const steps = [
```
tokland, line 8: As it's not calling any hook, let's use a non-useXyz name

src/webapp/utils.ts
```diff
@@ -0,0 +1,52 @@
...
+    return permission.noAccess()
+        ? "--"
+        : [permission.read ? "r" : "-", permission.write ? "w" : "-"].join("");
+}
+
+export function generateFullPermission(permissions: Permissions): string {
+    return `${convertPermissionToOctal(permissions.metadata)}${convertPermissionToOctal(
+        permissions.data
+    )}----`;
+}
```
tokland: A common advice is not to create a big utils file with unrelated functions. Here I'd say we have react.ts, permissions.ts and another one for parseSortField.

src/data/repositories/DataSetD2Api.ts
```diff
@@ -1,91 +1,102 @@
...
     private d2ApiCategoryOption: D2ApiCategoryOption;
-    constructor(private api: D2Api, private metadata: MetadataItem) {
+    private d2ApiConfig: D2ApiConfig;
+
+    constructor(private api: D2Api) {
         this.d2ApiCategoryOption = new D2ApiCategoryOption(this.api);
+        this.d2ApiConfig = new D2ApiConfig(this.api);
     }
 
     get(options: D2DataSetOptions): FutureData<Paginated<DataSet>> {
```
tokland: In general, we try to keep methods short and descriptive, and that's special relevant for public methods. In a glance, we should be able to tell what steps the public method is performing. We can abstract operations to instance methods.

src/data/repositories/DataSetD2Api.ts
```diff
@@ -1,91 +1,102 @@
...
+                ? [attributes.project.id]
+                : [attributes.createdByApp.id];
+            return apiToFuture(
+                this.api.models.dataSets.get({
+                    pageSize: options.paging.pageSize,
+                    page: options.paging.page,
+                    filter: {
+                        "attributeValues.attribute.id": { in: attributesToFilter },
+                        "attributeValues.value": this.isProjectFilterPresent(options)
+                            ? { in: [...this.getOrThrow(options.filters.projectsIds)] }
```
tokland: I get the idea, you're trying to be declarative on the conditional. The problem is that TS is then unaware of the check and you have to add the hacky getOrThrow. One simple solution is to use the plain condition: `"attributeValues.value": options.filters.projectsIds`, and then TS knows about it. Whenever the predicate is really complex and we need a function, we can check how to do it.

src/data/repositories/DataSetD2Api.ts
```diff
@@ -1,91 +1,102 @@
...
+                        ...(options.includeOrgUnits
+                            ? {
+                                  organisationUnits: {
+                                      id: true,
+                                      path: true,
+                                      displayName: true,
+                                  },
+                              }
+                            : {}),
+                    },
```
tokland, line 47: Conditional fields will bring problems. If you check the response type, organisationUnits won't be nice to work with. It can be done with generics, but typically is better to keep it simple. 

As a rule of thumb: If we need a method that does not get orgunits, then we have a separate method and a separate dataSet type (without orgUnits). Setting empty fields is a code smell.

src/data/repositories/DataSetD2Api.ts
```diff
@@ -135,57 +147,32 @@ export class DataSetD2Api {
...
+        projects: Project[],
+        attributes: D2Config["attributes"]
     ): DataSet {
         const projectAttributeId = d2DataSet.attributeValues.find(
-            attribute => attribute.attribute.id === this.metadata.attributes.project.id
+            attribute => attribute.attribute.id === attributes.project.id
         )?.value;
         const projectDetails = projects.find(project => project.id === projectAttributeId);
         const dataElementGroups = this.buildDataElementsGroupsCodes(d2DataSet);
         return DataSet.create({
```
tokland, line 209: I'd add a blank before the return to let the code breath

src/data/repositories/DataSetD2Api.ts
```diff
@@ -248,6 +240,18 @@ export class DataSetD2Api {
...
+    private convertPermissionToOctal(permission: Permission): OctalNotationPermission {
+        return permission.noAccess()
+            ? "--"
+            : [permission.read ? "r" : "-", permission.write ? "w" : "-"].join("");
+    }
+
+    generateFullPermission(permissions: Permissions): OctalNotationPermission {
+        return `${this.convertPermissionToOctal(
+            permissions.metadata
+        )}${this.convertPermissionToOctal(permissions.data)}----`;
```
tokland: I think this will be more clear: ` [a,b,c].join("")`

src/webapp/routes.ts
```diff
@@ -0,0 +1,37 @@
...
+
+    const navigateTo = React.useCallback(
+        <T extends RouteKey>(route: T, params?: RouteParams<T>) => {
+            const path = (routes[route] as any)(params);
+            history.push(path);
+        },
+        [history]
+    );
+
+    return { navigateTo };
```
tokland: Not important, but we plan to add more methods here? if we are sure we are not, then return the plain object.

src/webapp/routes.ts
```diff
@@ -0,0 +1,37 @@
...
+export function useNavigateTo() {
+    const history = useHistory();
+
+    const navigateTo = React.useCallback(
+        <T extends RouteKey>(route: T, params?: RouteParams<T>) => {
+            const path = (routes[route] as any)(params);
+            history.push(path);
+        },
+        [history]
+    );
```
tokland, line 32: Nice, very close. There is only an edge case, when you try to use it like this: `navigateTo("editDataSets")`, right? TS won't warn about the missing params. In this scenario, the solution is using `...args` somehow.  I think that covers all cases (confirm):

```typescript
const navigateTo = React.useCallback(
    <Key extends RouteKey>(route: Key, ...args: Parameters<(typeof routes)[Key]>) => {
        const path = (routes[route] as any)(...args);
        history.push(path);
    },
    [history]
);
```

tokland (commented):
Very close, just some details:

MiquelAdell (approved):
tested against gorstraining and it works
thanks!

src/data/repositories/DataSetD2Api.ts
```diff
@@ -33,40 +33,85 @@ export class DataSetD2Api {
...
+    ): FutureData<Paginated<DataSet>> {
+        return apiToFuture(
+            this.api.models.dataSets.get({
+                pageSize: options.paging.pageSize,
+                page: options.paging.page,
+                filter: {
+                    "attributeValues.attribute.id": { in: attributesToFilter },
+                    "attributeValues.value": options.filters.projectsIds
+                        ? { in: [...this.getOrThrow(options.filters.projectsIds)] }
+                        : { eq: "true" },
```
tokland: This looks confusing to me. We have a filter id/value (that's a problem with DHIS2, as we already saw), and the filter is used to both filter by projectIds and something boolean (I guess, createdByApp). Not sure the best approach, but I think this needs some refactoring, first to be more declarative (join the key/value that you want to filter, in the options), and second how to manage the DHIS2 part.

src/data/repositories/DataSetD2Api.ts
```diff
@@ -33,40 +33,85 @@ export class DataSetD2Api {
...
+                    identifiable: { token: options.filters.search },
+                    id: { in: options.filters.ids },
+                },
+                fields: {
```
tokland: a lot of repeated code with the previous method, we can abstract it (if it's conceptually the same)

src/data/repositories/DataSetD2Repository.ts
```diff
@@ -40,11 +40,10 @@ export class DataSetD2Repository implements DataSetRepository {
     }
 
     private getDataSets(page: number, pageSize: number) {
-        return this.d2DataSetApi.get({
+        return this.d2DataSetApi.getWithOrgUnits({
             paging: { page, pageSize },
             filters: {},
             sorting: { field: "name", order: "asc" },
-            includeOrgUnits: true,
```
tokland: `type D2DataSetOptions = GetDataSetOptions & { includeOrgUnits?: boolean };`

is `includeOrgUnits` used, after the refactor?

tokland (commented):
In-line:

PR #470, opened 2024-11-26
Add setup screen

eperedo:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/8696ewfj9

### :memo: Implementation

I'm going to create a PR for every step in the wizard in order to avoid a long and complex testing process.

[Project-Configuration-app_create_dataset.webm](https://github.com/user-attachments/assets/483fb94a-62cf-430e-9868-cbfbae1f8ce0)


MiquelAdell (approved):

i18n/es.po
```diff
@@ -183,21 +201,55 @@ msgstr ""
...
 msgid "The dataSet is finished. Press the button Save to save the data"
 msgstr ""
 
+msgid "Save"
+msgstr ""
+
+msgid "and {{number}} more."
+msgstr ""
+
+#, fuzzy
```
tokland: Remove fuzzy comments (check all the files)

src/data/repositories/DataSetD2Repository.ts
```diff
@@ -183,28 +203,25 @@ export class DataSetD2Repository implements DataSetRepository {
...
-                    attribute: { id: projectAttributeId },
-                    value: dataSet.project.id,
-                },
-            ];
-        }
+        // if (!dataSet.project) return existingAttributes || [];
+        // const projectAttributeId = attributes.project.id;
+        // const projectAttribute = existingAttributes?.find(
+        //     attribute => attribute.attribute.id === projectAttributeId
+        // );
```
tokland: remove commented code

src/data/repositories/DataSetD2Repository.ts
```diff
@@ -183,28 +203,25 @@ export class DataSetD2Repository implements DataSetRepository {
...
+
+        const pa = { attribute: { id: attributes.project.id }, value: dataSet.project?.id };
```
tokland: pa -> more declarative name

src/data/repositories/DataSetD2Repository.ts
```diff
@@ -183,28 +203,25 @@ export class DataSetD2Repository implements DataSetRepository {
...
+        const createdByAttribute = { attribute: { id: attributes.createdByApp.id }, value: "true" };
+
+        const attributesToSave = _([pa, createdByAttribute])
+            .compactMap(attribute => (attribute.value ? attribute : undefined))
```
tokland: Maybe using `filter`?

src/data/repositories/DataSetD2Repository.ts
```diff
@@ -183,28 +203,25 @@ export class DataSetD2Repository implements DataSetRepository {
...
+            .value();
+
+        const filteredExisting =
+            existingAttributes?.filter(
+                attr => !attributesToSave.some(save => save.attribute.id === attr.attribute.id)
+            ) || [];
+
+        // Combinar `filteredExisting` con `toSave`
```
tokland: temporal comment?

src/data/repositories/OrgUnitD2Repository.ts
```diff
@@ -0,0 +1,38 @@
...
+import { Id } from "$/domain/entities/Ref";
+import { Future, FutureData } from "$/domain/entities/generic/Future";
+import { OrgUnitRepository } from "$/domain/repositories/OrgUnitRepository";
+import { chunkRequest } from "$/data/utils";
+
+export class OrgUnitD2Repository implements OrgUnitRepository {
+    constructor(private api: D2Api) {}
+
+    getByIds(ids: Id[]): FutureData<OrgUnit[]> {
+        if (ids.length === 0) return Future.success([]);
```
tokland: Not that this is wrong, but typically we shouldn't need it, the processing logic should support empty get (in this case, chunkRequest -> not sure if it already does)

src/data/repositories/OrgUnitD2Repository.ts
```diff
@@ -0,0 +1,38 @@
...
+            return apiToFuture(
+                this.api.models.organisationUnits.get({
+                    fields: { id: true, displayName: true, path: true },
+                    filter: { id: { in: idsToFetch } },
+                    paging: false,
+                })
+            ).map(response => response.objects);
+        });
+
+        return $requests.map(response => {
```
tokland: minor, but let's be more declarative, for example: `$requests -> d2OrgsUnits$`, `response -> d2OrgUnit`

src/data/repositories/OrgUnitD2Repository.ts
```diff
@@ -0,0 +1,38 @@
...
+            // const allRecords = response.flatMap(r => r);
```
tokland: temporal comment?

src/data/repositories/OrgUnitTestRepository.ts
```diff
@@ -0,0 +1,9 @@
+import { OrgUnit } from "$/domain/entities/DataSet";
+import { FutureData } from "$/domain/entities/generic/Future";
+import { OrgUnitRepository } from "$/domain/repositories/OrgUnitRepository";
+
+export class OrgUnitTestRepository implements OrgUnitRepository {
+    getByIds(): FutureData<OrgUnit[]> {
+        throw new Error("Method not implemented.");
+    }
```
tokland, line 8: So we don't have time for tests?

src/data/repositories/ProjectD2Repository.ts
```diff
@@ -21,10 +21,52 @@ export class ProjectD2Repository implements ProjectRepository {
...
+                        lastUpdated: d2CategoryOption.lastUpdated,
+                        isOpen: this.isProjectOpen(
+                            d2CategoryOption.startDate,
+                            d2CategoryOption.endDate
+                        ),
+                    });
+                });
+            });
+        });
+    }
```
tokland, line 34: Too long, specially for a public method. The block is doing 3 things, the first one (getCategories) is already abstracted, we can abstract the other two. And then it would be clear at a glance what the method is doing.

src/data/repositories/ProjectD2Repository.ts
```diff
@@ -21,10 +21,52 @@ export class ProjectD2Repository implements ProjectRepository {
...
+
     getAll(): FutureData<Project[]> {
         return this.getAllProjects(1, []);
     }
 
+    private isProjectOpen(date1: ISODateString, date2: ISODateString): boolean {
+        if (!date1 || !date2) return false;
```
tokland, line 70: TS-smell. the type says we do have a ISODateString, but we need to check if empty. So probably the type is `Maybe<IsoDateString>`. I see this comes from a `d2CategoryOption.startDate`. This, I guess, is a shortcoming of d2-api, you ask for a field but this field may not be set so it's empty in the response. Until we have support for that in d2-api, cast explicitly as soon as possible so types are exact.

src/domain/entities/DataSet.ts
```diff
@@ -32,6 +39,32 @@ export type CoreCompetency = { id: Id; name: string; code: string };
...
+
+        return errors.length === 0 ? Either.success(this) : Either.error(errors);
+    }
+
+    updateProject(project: Maybe<Project>): DataSet {
+        const name = project ? `${project.name} DataSet` : "";
+        return this._update({ project, name });
+    }
+
+    update(fieldName: keyof DataSet, value: string | number | boolean): DataSet {
```
tokland: This is not type-safe. The usual: when you need to reference a type across arguments/output type, you create a type argument: `update<K extends keyof DataSet>(fieldName: K, value: DataSet[K]): DataSet`

src/domain/entities/DataSet.ts
```diff
@@ -40,6 +73,15 @@ export class DataSet extends Struct<DataSetAttrs>() {
         return DataSet.create({ ...this, orgUnits });
     }
 
+    static buildOrgUnitsFromPaths(paths: string[]): OrgUnit[] {
+        const orgUnits = paths.map(path => ({
+            id: _(path.split("/")).last() || "",
+            name: path,
+            path: path.split("/").slice(1),
```
tokland: In our domain, a path is an array of IDs, right? But these seem DHIS2 paths, so this processing should not be in the domain layer.

src/domain/entities/generic/Error.ts
```diff
@@ -0,0 +1,44 @@
...
+}
+
+export function getErrors<T>(errors: ValidationError<T>[]): string[] {
+    return errors
+        .map(error => {
+            return error.errors.map(err =>
+                validationErrorMessages[err](error.property as string, error.value)
+            );
+        })
+        .flat();
```
tokland: flatMap

src/domain/entities/generic/Error.ts
```diff
@@ -0,0 +1,44 @@
...
+        })
+        .flat()
+        .join("\n");
+}
+
+export function getErrors<T>(errors: ValidationError<T>[]): string[] {
+    return errors
+        .map(error => {
+            return error.errors.map(err =>
+                validationErrorMessages[err](error.property as string, error.value)
```
tokland: Instead of these two `as string` you can define the type below: `property: keyof T & string`

src/domain/usecases/SaveDataSetUseCase.ts
```diff
@@ -0,0 +1,39 @@
...
+                shortName: this.truncateValue(dataSet.name),
+            });
+
+            const result = dataSetToSave.validateSetup();
+            if (result.isError()) {
+                const errors = getErrors(result.value.error);
+                return Future.error(new Error(errors.join("\n")));
+            }
+
+            return this.dataSetRepository.save([dataSetToSave]);
```
tokland: Not sure about this. For me, `this.dataSetRepository.save([dataSet])` should take care of get the existing + merge + save. Unless the domain needs to interact somehow with the existing dataSet, but I think that's not the case. And then the validation would be done for dataSet.

src/domain/usecases/SaveDataSetUseCase.ts
```diff
@@ -0,0 +1,39 @@
...
+    private getDataSetById(id: string): FutureData<Maybe<DataSet>> {
+        return this.dataSetRepository.getByIds([id]).map(dataSets => {
+            return _(dataSets).first();
+        });
+    }
+
+    private truncateValue(input: string): string {
+        const targetLength = 50;
+        return input.length > targetLength ? input.slice(0, targetLength) : input;
+    }
```
tokland, line 18: It seems that `dataSet.shortValue` can be a `get` property in the entity, right? having manually to derive it has problems: we have to remember to do it when necessary.

src/domain/usecases/ValidateNameUseCase.ts
```diff
@@ -0,0 +1,18 @@
...
+        return this.dataSetRepository.getByName(options.name).map(dataSets => {
+            return dataSets.some(
+                dataSet =>
+                    dataSet.id !== options.dataSetId &&
+                    dataSet.name.toLowerCase() === options.name.toLowerCase()
+            );
+        });
+    }
+}
+export type ValidateNameOptions = { name: string; dataSetId: Id };
```
tokland: name/dataSetId, both should contain "dataSet" (or none). I'd change all the naming of the use case/types: ValidateNameUseCase -> ValidateDataSetNameUseCase.

src/webapp/components/dataset-wizard/DataSetWizard.tsx
```diff
@@ -17,22 +26,77 @@ const useStyles = makeStyles((theme: Theme) =>
...
+        (currentStep: WizardStep) => {
+            if (validationInProgressOrError)
+                return Promise.resolve(["Validation name in progress"]);
+            if (currentStep.key === "setup") {
+                const result = dataSet.validateSetup();
+                if (result.isError()) {
+                    return Promise.resolve(getErrors(result.value.error));
+                }
+            }
+            return Promise.resolve([]);
```
tokland: if/else if/else

src/webapp/components/dataset-wizard/DataSetWizard.tsx
```diff
@@ -17,22 +26,77 @@ const useStyles = makeStyles((theme: Theme) =>
...
+        },
+        [dataSet, validationInProgressOrError]
+    );
```
tokland: Too long. The validation stuff can be moved to a custom hook.

src/webapp/pages/register-dataset/RegisterDataSetPage.tsx
```diff
@@ -1,10 +1,103 @@
...
+            permissions: {
+                data: Permission.create({ read: false, write: false }),
+                metadata: Permission.create({ read: false, write: false }),
+            },
+            project: undefined,
+            shortName: "",
+            expiryDays: 0,
+            openFuturePeriods: 0,
+            notifyUser: false,
+        });
```
tokland: Maybe move to `DataSet.initial()` ?

src/webapp/pages/register-dataset/RegisterDataSetPage.tsx
```diff
@@ -1,10 +1,103 @@
...
+                snackbar.error(error.message);
+                setStatus("error");
+            }
+        );
+    }, [compositionRoot.dataSets.getByIds, id, snackbar]);
+
+    return { dataSet, status, updateDataSet };
+}
+
+export type HttpStatus = "idle" | "loading" | "finished" | "error";
```
tokland: not important, but it looks a bit weird :) Maybe RequestStatus, LoadingStatus or similar.

tokland (commented):
Some in-line comments:

src/data/repositories/OrgUnitTestRepository.ts
eperedo: I'm adding tests for the share step, where the logic is more complex than just validating required values.

src/data/repositories/ProjectD2Repository.ts
```diff
@@ -23,32 +27,8 @@ export class ProjectD2Repository implements ProjectRepository {
...
-                        id: d2CategoryOption.id,
-                        name: d2CategoryOption.displayName,
-                        lastUpdated: d2CategoryOption.lastUpdated,
-                        isOpen: this.isProjectOpen(
-                            d2CategoryOption.startDate,
-                            d2CategoryOption.endDate
-                        ),
-                    });
-                });
+            return this.getCategoryOptionsByCode(categories.project.code).map(d2Response => {
```
tokland, line 30: Typically, we'd make getCategoryOptionsByCode return directly the categoryOptions, no need for the full response. It's a way of narrowing scopes.

tokland (approved):
All good, code-wise, just a very minor thing, no re-review needed.

PR #471, opened 2024-12-02
add indicators screen

eperedo:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/8696ewg1c

### :memo: Implementation

- [x] outputs indicators are saved as `dataElements` within dataSet and sections
- [x] outcomes indicators are saved as `indicators` within dataSet and sections
- [x] outcomes indicators can include extra dataElements when the indicator (DHIS) include references to dataElements in: numerator, denominator or code.

[Project-Configuration-app_indicators.webm](https://github.com/user-attachments/assets/b42c0396-a2e6-487e-907a-273e597b6547)

Using [PR 470](https://github.com/EyeSeeTea/dataset-configuration/pull/470) as base branch to facilitate code review

#8696ewg1c


MiquelAdell (approved):

src/data/repositories/CoreCompetencyD2Repository.ts
```diff
@@ -0,0 +1,50 @@
...
+                                  code: d2DataElementGroup.code,
+                                  name: d2DataElementGroup.displayName,
+                              })
+                          )
+                      )
+                    : Future.error(
+                          new Error(i18n.t(`Core competency group not found: ${coreCompetencyId}`))
+                      );
+            });
+        });
```
tokland, line 23: Break in 2 declarative methods (move the existence check to the first one)

src/data/repositories/D2ApiConfig.ts
```diff
@@ -47,30 +75,92 @@ export class D2ApiConfig {
...
                     ),
+                    theme: getOrThrow(
+                        d2Response.dataElementGroupSets,
+                        metadataCodes.dataElementGroupSets.theme
+                    ),
+                    status: getOrThrow(
+                        d2Response.dataElementGroupSets,
+                        metadataCodes.dataElementGroupSets.status
+                    ),
+                },
```
tokland: Writing one-liner is not generally a goal, but for this kind of code, the structure will look much nicer if are able to have one-liners. For that, we create function helpers that abstract the repetitions and local variables (shortening the name, only if necessary). For example, the previous three calls can be captured with something like that:

```
            const { dataElementGroupSets } = metadataCodes;
            const getGroupSet = (groupSetCode: string) =>
                getOrThrow(d2Response.dataElementGroupSets, groupSetCode);
```

And now is used like this:

```
            coreCompetency: getGroupSet(dataElementGroupSets.coreCompetency),
```

Same for the rest of gets (name of the helper is just a proposal)

src/data/repositories/D2ApiConfig.ts
```diff
@@ -6,11 +6,27 @@ import { apiToFuture } from "$/data/api-futures";
 
 export const metadataCodes = {
```
tokland: The filename looks a bit confusing to me, as by D2Api we think of the library. Maybe D2Metadata?

src/data/repositories/DataElementD2Repository.ts
```diff
@@ -0,0 +1,45 @@
...
+                    disaggregation: d2DataElement.categoryCombo
+                        ? {
+                              id: d2DataElement.categoryCombo.id,
+                              name: d2DataElement.categoryCombo.displayName,
+                          }
+                        : undefined,
+                };
+            });
+        });
+    }
```
tokland, line 21: As usual, let's do a get + build (probably removing the initial empty check) so the method looks very declarative.

src/data/repositories/DataSetD2Repository.ts
```diff
@@ -101,7 +103,7 @@ export class DataSetD2Repository implements DataSetRepository {
         const ids = dataSets.map(dataSet => dataSet.id);
 
         return this.d2DataSetApi.getConfig().flatMap(config => {
```
tokland, line 105: save method too long, split steps

src/data/repositories/DataSetD2Repository.ts
```diff
@@ -160,6 +165,115 @@ export class DataSetD2Repository implements DataSetRepository {
...
+            return this.deleteSections(sectionsToDelete).flatMap(() => {
+                const sectionsToSave = sectionsActions.flatMap(
+                    sectionAction => sectionAction.toSave
+                );
+                return apiToFuture(this.api.metadata.post({ sections: sectionsToSave })).flatMap(
+                    sectionResponse => {
+                        const allErrors = this.extractErrorsFromResponse(sectionResponse);
+                        if (allErrors.length > 0)
+                            return Future.error(new Error(allErrors.join("\n")));
+                        return Future.success(undefined);
```
tokland, line 211: this pattern will be very common, make a function that cares of that with a single call (example: runMetadata in CPR).

src/data/repositories/DataSetD2Repository.ts
```diff
@@ -221,4 +366,27 @@ export class DataSetD2Repository implements DataSetRepository {
...
+    id: string;
+    code: string;
+    dataSet: Ref;
+    name: string;
+    greyedFields: Array<{ dataElement: Ref; categoryOptionCombo: Ref }>;
+    dataElements: Ref[];
+    indicators: Ref[];
+};
+
+const indicatorTypeLabel = { outcomes: "Outcomes", outputs: "Outputs" };
```
tokland: very minor, but best practices: `const indicatorTypeLabel: Record<IndicatorAttrs["type"], string> = {`

src/domain/entities/DataElement.ts
```diff
@@ -0,0 +1,4 @@
+import { Id, NamedRef } from "$/domain/entities/Ref";
+import { Maybe } from "$/utils/ts-utils";
+
+export type DataElement = { id: Id; name: string; code: string; disaggregation: Maybe<NamedRef> };
```
tokland: subjective, let's use multi-line type for clarity.

src/domain/entities/DataSet.ts
```diff
@@ -39,6 +41,11 @@ export type CoreCompetency = { id: Id; name: string; code: string };
 export type DataSetList = Pick<DataSetAttrs, "id" | "name" | "lastUpdated" | "permissions">;
 
 export class DataSet extends Struct<DataSetAttrs>() {
+    validate(): Either<ValidationError<DataSet>[], DataSet> {
+        const allErrors = this.getValidationErrors();
+        return allErrors.length === 0 ? Either.success(this) : Either.error(allErrors);
+    }
```
tokland, line 46: just as a note, an Either here is no strictly necessary, the return type could be `ValidationError<DataSet>[] | undefined`. But it's subjective if this is more declarative.

src/domain/entities/DataSet.ts
```diff
@@ -82,6 +77,31 @@ export class DataSet extends Struct<DataSetAttrs>() {
         return DataSet.create({ ...this, orgUnits });
     }
 
+    setIndicators(indicators: Indicator[]): DataSet {
+        return this._update({ indicators });
+    }
+
+    validateIndicatorsStep(): Either<ValidationError<DataSet>[], DataSet> {
```
tokland: same

src/domain/entities/DataSet.ts
```diff
@@ -92,6 +112,28 @@ export class DataSet extends Struct<DataSetAttrs>() {
         return dataSets.map(dataSet => dataSet.shortName).join(separator);
     }
 
+    private getValidationErrors(): ValidationError<DataSet>[] {
+        const setupErrors = this.buildSetupErrors();
+        const indicatorsErrors = this.validateIndicatorsStep().value.error || [];
```
tokland: [note] this usage is considered unorthodox (having error: never on a success value is kind of a hack)

src/domain/entities/DataSet.ts
```diff
@@ -82,6 +77,31 @@ export class DataSet extends Struct<DataSetAttrs>() {
...
+                      value: this.indicators,
+                  },
+              ]);
+    }
+
+    static buildOrgUnitsFromPaths(paths: string[]): OrgUnit[] {
+        const orgUnits = paths.map(path => ({
+            id: _(path.split("/")).last() || "",
+            name: path,
+            path: path.split("/").slice(1),
```
tokland: Ouch, I remember this is something we fixed in another PR, right, not having path:string in the domain, maybe I am repeating reviews because of the base branch?

src/domain/entities/Indicator.ts
```diff
@@ -0,0 +1,38 @@
...
+import { Maybe } from "$/utils/ts-utils";
+
+export type IndicatorAttrs = {
+    id: Id;
+    name: string;
+    code: string;
+    theme: string;
+    status: string;
+    type: "outputs" | "outcomes";
+    scope: "core" | "local" | "donor";
```
tokland: As this kind of types are usually used isolated, let's create an alias.

src/domain/entities/Settings.ts
```diff
@@ -0,0 +1,43 @@
...
+    };
+    outputEndDate: {
+        day: number;
+        month: number;
+    };
+    outputLastYearEndDate: {
+        units: string;
+        value: number;
+    };
+};
```
tokland: note: here we have kind of a mismatch. These are settings of things we consider to be data details (ex: dataElementGroupSets), so conceptually, should not be in the domain. 

But I understand you will want it configurable, so it will have UI so some link with the domain. 

It could be something we pass to the d2 repo in the constructor so it does not appear in the domain, but then the composition root should be recreated after UI changes. And the UI would be referencing some structure not in domain.  So if you have time and that worries you, check with Jorge.

src/domain/usecases/GetCoreCompetencyUseCase.ts
```diff
@@ -0,0 +1,11 @@
+import { CoreCompetency } from "$/domain/entities/DataSet";
+import { FutureData } from "$/domain/entities/generic/Future";
+import { CoreCompetencyRepository } from "$/domain/repositories/CoreCompetencyRepository";
+
+export class GetCoreCompetencyUseCase {
```
tokland: I see that at least this use case has no usage in the app. Typically we won't do that, only create it when known to be used, but of course, you know more context and you may be sure it will be used isolatedly in the UI.

src/domain/usecases/ValidateDataSetNameUseCase.ts
```diff
@@ -1,18 +1,14 @@
...
 
     execute(options: ValidateDataSetNameOptions): FutureData<boolean> {
-        return this.dataSetRepository.getByName(options.name).map(dataSets => {
-            return dataSets.some(
-                dataSet =>
-                    dataSet.id !== options.dataSetId &&
-                    dataSet.name.toLowerCase() === options.name.toLowerCase()
-            );
-        });
+        return this.dataSetUtils.dataSetNameExists(options);
```
tokland: `dataSetNameExists` typically we want a verb in a method.

src/webapp/components/dataset-wizard/FilterIndicators.tsx
```diff
@@ -6,20 +6,25 @@ import FilterListIcon from "@material-ui/icons/FilterList";
...
 export type FilterIndicatorsProps = {
-    coreCompetencies: string[];
-    coreValue: string;
+    coreCompetencies: CoreCompetency[];
+    coreValues: string[];
     groups: string[];
+    group: string;
     onClose: () => void;
-    onFilterChange: (scope: string, type: FilterType) => void;
+    onFilterChange: (scope: ChipItem | ChipItem[], type: FilterType) => void;
```
tokland: typically, we try to avoid this JS (T | T[]) acceptance (we accept T[] and create another function that wrap the single case)

src/webapp/components/dataset-wizard/IndicatorsDataSet.tsx
```diff
@@ -23,78 +33,133 @@ export type IndicatorsColumns = {
...
-                setCore(value);
-                break;
-        }
-    }, []);
+    const updateFilter = React.useCallback(
+        (value: ChipItem | ChipItem[], filterType: FilterType) => {
+            const isArray = Array.isArray(value);
+            switch (filterType) {
+                case "scope":
+                    if (!isArray) {
```
tokland: ... and then we can avoid this ugly code :)

src/webapp/pages/register-dataset/RegisterDataSetPage.tsx
```diff
@@ -26,57 +28,60 @@ const RegisterDataSetPage_ = () => {
...
+export function useGetDataSetSettings(props: { id: Id }) {
     const { id } = props;
     const { compositionRoot } = useAppContext();
     const snackbar = useSnackbar();
     const [status, setStatus] = React.useState<LoadingStatus>("idle");
-    const [dataSet, updateDataSet] = React.useState<DataSet>(() => {
-        return DataSet.initial(getUid(new Date().getTime().toString()));
-    });
+    const [dataSet, updateDataSet] = React.useState<DataSet>(
+        DataSet.initial(getUid(new Date().getTime().toString()))
```
tokland: abstract to get RandomUid or similar

tokland (commented):
In-line comments:

src/domain/entities/Settings.ts
eperedo: I removed this entity since I initially planned to include the settings form in this PR but later realized it would be better to handle it in a separate one

src/domain/entities/DataSet.ts
eperedo: My bad. I forgot to remove it after the merge.

src/domain/entities/Settings.ts
tokland: Ok. Anyway, you'll face the mismatch later, we'll see what's the best approach.

src/domain/entities/DataSet.ts
```diff
@@ -41,9 +41,8 @@ export type CoreCompetency = { id: Id; name: string; code: string };
 export type DataSetList = Pick<DataSetAttrs, "id" | "name" | "lastUpdated" | "permissions">;
 
 export class DataSet extends Struct<DataSetAttrs>() {
-    validate(): Either<ValidationError<DataSet>[], DataSet> {
-        const allErrors = this.getValidationErrors();
-        return allErrors.length === 0 ? Either.success(this) : Either.error(allErrors);
+    validate(): ValidationError<DataSet>[] {
```
tokland, line 44: minor, no need to change, just a general note:  Using `ValidationError<DataSet>[]` constrains the return type to always be an array. For a more flexible approach, consider `ValidationErrors<DataSet>` instead, allowing it to represent any structure. This way, the type can be reshaped in the future without requiring changes to all signatures where it is passed, focusing only on the code that interacts with its internals.

src/domain/entities/Indicator.ts
```diff
@@ -12,7 +12,7 @@ export type IndicatorAttrs = {
     theme: string;
     status: string;
     type: "outputs" | "outcomes";
-    scope: "core" | "local" | "donor";
+    scope: ScopeType;
```
tokland: To keep the naming consistent: `Scope` or `IndicatorScope`

src/utils/uid.ts
```diff
@@ -45,6 +45,9 @@ export function generateUid(): string {
     return randomChars;
 }
 
+// @ts-ignore
+window.generateUid = generateUid;
+
```
tokland: Do we need this?

src/webapp/components/dataset-wizard/FilterIndicators.tsx
```diff
@@ -135,7 +138,7 @@ export const FilterIndicators = React.memo((props: FilterIndicatorsProps) => {
 export type ChipFilterProps = {
     items: ChipItem[];
     label: string;
-    onChange: (item: ChipItem | ChipItem[]) => void;
+    onChange: (item: ChipItem[]) => void;
     value: string | string[];
```
tokland: sorry, I hadn't seen that. Can we follow the same idea for value, to be only `string[]`? 

The general idea is that, when using TS code in our control, we never have to call functions like isArray, isObject, isString, etc.

(When we really need to have this kind of distinctions, in other scenarios, we use type-safe discriminating union types)

src/data/repositories/D2ApiMetadata.ts
```diff
@@ -65,60 +65,63 @@ export class D2ApiConfig {
...
                 categories: {
-                    project: getOrThrow(d2Response.categories, metadataCodes.categories.project),
+                    project: getOrThrowMetadata("categories", metadataCodes.categories.project),
                 },
                 dataElementGroupSets: {
-                    coreCompetency: getOrThrow(
-                        d2Response.dataElementGroupSets,
+                    coreCompetency: getOrThrowMetadata(
+                        "dataElementGroupSets",
                         metadataCodes.dataElementGroupSets.coreCompetency
```
tokland, line 79: We were not able to get single lines, but yes, it may hinder readability if we push it too much. let's keep as is.

(other strategies: move to separate function to reduce indentation level, create shorter name for helper functions, rename to shorter names the variables in the block)

tokland (commented):
Some in-line comments:

tokland (approved):
All good by me, code-wise

PR #478, opened 2025-02-09
add period dates action. script to migrate existing data

eperedo:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/8695uy4nz

### :memo: Implementation

- [x] create new attribute for period dates (metadata in issue)
- [x] add action to dataSet list
- [x] script to migrate existing data to new attribute

```bash
 npx ts-node src/scripts/migrate-dataset-period-dates.ts -u 'http://localhost:8080' --username 'username' --password 'password'
```

[Project-Configuration-app-period_dates.webm](https://github.com/user-attachments/assets/66acd435-957e-4ab5-894f-f667ae3825f5)

#8695uy4nz

### DIFF
--- a/src/data/utils.ts
+++ b/src/data/utils.ts
@@ -1,11 +1,12 @@
 import { FutureData } from "$/domain/entities/generic/Future";
-import { Id } from "$/domain/entities/Ref";
+import { ISODateString, Id } from "$/domain/entities/Ref";
 import { Future } from "$/domain/entities/generic/Future";
 import _ from "$/domain/entities/generic/Collection";
 import { MetadataResponse } from "@eyeseetea/d2-api/api";
 import { CancelableResponse } from "@eyeseetea/d2-api";
 import { apiToFuture } from "$/data/api-futures";
 import { Category } from "$/domain/entities/Category";
+import { Maybe } from "$/utils/ts-utils";
 
 export function chunkRequest<Res>(
     ids: Id[],
@@ -23,8 +24,6 @@ export function chunkRequest<Res>(
 }
 
 export function getErrorFromResponse(res: MetadataResponse): string {
-    console.debug(JSON.stringify(res, null, 4));
-
     const errorsMessages = res.typeReports
         .flatMap(typeReport => typeReport.objectReports)
         .flatMap(objectReport => objectReport.errorReports)
@@ -54,6 +53,31 @@ export function convertToCategories(d2Categories: D2Category[]): Category[] {
     }));
 }
 
+export function convertAttributeValueToDate(
+    dateString: Maybe<D2AttributeDateValue>
+): ISODateString {
+    if (!dateString) return "";
+
+    if (!/^\d{8}$/.test(dateString)) {
+        console.warn("invalid format, expected 8 digits", dateString);
+        return "";
+    }
+
+    const year = parseInt(dateString.substring(0, 4), 10);
+    const month = parseInt(dateString.substring(4, 6), 10) - 1;
+    const day = parseInt(dateString.substring(6, 8), 10);
+
+    const date = new Date(year, month, day, 0, 0, 0);
+
+    if (date.getFullYear() !== year || date.getMonth() !== month || date.getDate() !== day) {
+        console.warn("Invalid date format. Expected YYYYMMDD");
+        return "";
+    }
+
+    return date.toISOString();
+}
+
 export type D2NamedRef = { id: Id; displayName: string };
 export type D2Category = D2NamedRef & { categoryOptions: D2NamedRef[] };
 export type D2CategoryCombo = D2NamedRef & { categories: D2Category[] };
+type D2AttributeDateValue = string; // format: YYYYMMDD
--- a/src/domain/entities/DataSetPeriodDate.ts
+++ b/src/domain/entities/DataSetPeriodDate.ts
@@ -1,0 +1,10 @@
+import { PeriodDate } from "$/domain/entities/PeriodDate";
+import { Id } from "$/domain/entities/Ref";
+
+export type DataSetPeriodDate = {
+    id: Id;
+    name: string;
+    periodDate: PeriodDate;
+    outcomeDate: PeriodDate;
+    outputDate: PeriodDate;
+};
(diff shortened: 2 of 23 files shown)


MiquelAdell:
@tokland I merged this PR in a rush but, whenever you have time, I would like to provide a code review anyway and we can have a different PR with the changes if needed.

Thanks!

src/webapp/components/dataset-periods/DataSetPeriodDates.tsx
```diff
@@ -0,0 +1,186 @@
...
+import { PeriodDate, PeriodDetailsAttrs } from "$/domain/entities/PeriodDate";
+import i18n from "$/utils/i18n";
+
+export type DataSetPeriodDatesProps = {
+    dataSetIds: Id[];
+    onCancel: () => void;
+    onSave: (periodDate: PeriodDate) => void;
+};
+
+function useGetYears(props: { period: PeriodDate }) {
```
tokland, line 18: Typically, we put props+component+hooks/helper functions

src/webapp/components/dataset-periods/DataSetPeriodDates.tsx
```diff
@@ -0,0 +1,186 @@
...
+                                    minDate={periodDate.startDate}
+                                    className="datepicker"
+                                    format="yyyy-MM-DD"
+                                />
+                            </PeriodDateItem>
+                        );
+                    })}
+                </PeriodsContainer>
+            )}
+        </ConfirmationDialog>
```
tokland, line 162: We may add a blank line between non-single-line components at the same indentation level (for example 129/130)

src/webapp/components/dataset-periods/DataSetPeriodDates.tsx
```diff
@@ -0,0 +1,186 @@
...
+                                    minDate={periodDate.startDate}
+                                    format="yyyy-MM-DD"
+                                />
+                                <DatePicker
+                                    value={periodDetail.endDate}
+                                    onChange={value => updatePeriod(periodDetail, value, "endDate")}
+                                    label={i18n.t("End date of data input")}
+                                    minDate={periodDate.startDate}
+                                    className="datepicker"
+                                    format="yyyy-MM-DD"
```
tokland, line 155: repeated prop format. Either a variable or a wrapper component that adds common props.

src/webapp/components/dataset-periods/DataSetPeriodDates.tsx
```diff
@@ -0,0 +1,186 @@
...
+    const onSavePeriods = () => {
+        if (!periodDate) return;
+        const periods = periodDate.periods.length === 0 ? periodsByYear : periodDate.periods;
+        onSave(periodDate.updatedPeriods(periods));
+    };
+
+    const onUpdatePeriodDate = (date: string, fieldName: "startDate" | "endDate") => {
+        const newValues = periodDate?.setDates(date, fieldName);
+        setPeriodDate(newValues);
+    };
```
tokland, line 99: Move to custom hook

src/scripts/migrate-dataset-period-dates.ts
```diff
@@ -0,0 +1,51 @@
...
+            const api = new D2Api({ baseUrl: args.url, auth: auth });
+            const configRepository = new ConfigD2Repository(api);
+            const config = await configRepository.get().toPromise();
+            const compositionRoot = getWebappCompositionRoot(api, config);
+            compositionRoot.dataSets.migratePeriodDates.execute().run(
+                response => {
+                    console.debug(JSON.stringify(response, null, 4));
+                },
+                error => {
+                    console.error(error);
```
tokland, line 42: [good practices] process.exit(1) on the error branch

src/domain/entities/PeriodDate.ts
```diff
@@ -0,0 +1,52 @@
...
+        if (!startDate || !endDate) return [];
+
+        const startYear = new Date(startDate).getFullYear();
+        const endYear = new Date(endDate).getFullYear();
+
+        return Array.from({ length: endYear - startYear + 1 }, (_, index) => startYear + index);
+    }
+
+    setDates(date: string, fieldName: "startDate" | "endDate"): PeriodDate {
+        return this._update({ [fieldName]: date });
```
tokland, line 46: That's unfortunate, but TS does not enforce type-safety when building objects from dynamic keys:

https://github.com/microsoft/TypeScript/issues/13948

A 100% type-safe workaround is using: `const newAttrs = fromPairs([[fieldName, date]])` (fromPairs from ts-utils or EST lodash should both work)

src/data/utils.ts
```diff
@@ -54,6 +53,31 @@ export function convertToCategories(d2Categories: D2Category[]): Category[] {
     }));
 }
 
+export function convertAttributeValueToDate(
```
tokland, line 56: this is a very specific non-generic function, utils.ts probably is not be the best place.

src/domain/entities/DataSetPeriodDate.ts
```diff
@@ -0,0 +1,10 @@
+import { PeriodDate } from "$/domain/entities/PeriodDate";
+import { Id } from "$/domain/entities/Ref";
+
+export type DataSetPeriodDate = {
+    id: Id;
+    name: string;
+    periodDate: PeriodDate;
+    outcomeDate: PeriodDate;
+    outputDate: PeriodDate;
+};
```
tokland, line 10: [no need to change]

I see you created an entity+repo for DataSetPeriodDate. 

This is something we have been discussing lately, and I'd like to confirm with @xurxodev, but from our last chat about this, he stated that having different "flavors" or entities is not correct. I see two alternatives: 

- Add these 3 fields would be in DataSet. I think this is the most simple option. And then we simply improve the existing DataSet/DataSetRepository. You may then say that we typically don't need them and getting the info from the attribute is very costly (i don't think it's the case here). Then we'd see alternatives.

-Have a DataSetPeriodDate, but not as flavor, but as its own entity. So having `dataSetId: Id` + 3 fields. Typically we do this if the entity has sense in isolation. Here it's 1:1 relation, it does not seem a typical case for this.

tokland (commented):
[code-only review] Already merged, but still requested by @MiquelAdell. In-line comments:

PR #481, opened 2025-02-19
add grey fields and custom form

eperedo:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/8696ewgtm, https://app.clickup.com/t/8698186gt, https://app.clickup.com/t/8698c7crh, https://app.clickup.com/t/8698c76yq

### :memo: Implementation

- [x] Add grey fields screen
- [x] Generate custom form

https://github.com/user-attachments/assets/fb150a97-1eff-40e8-9fed-d8ce980d0fad

Custom Form

image

#8696ewgtm
#8698186gt


MiquelAdell (approved):
tested in gors training and working. Thanks!

README.md
```diff
@@ -70,6 +70,12 @@ $ yarn localize
 
 Check the example script, entry `"script-example"`in `package.json`->scripts and `src/scripts/example.ts`.
 
+Migrate output dates to period dates
+
+```bash
+ npx ts-node src/scripts/migrate-dataset-period-dates.ts -u 'http://localhost:8080' --username 'username' --password 'password'
```
tokland: Let's use "yarn run-script" to abstract the operation, in case we change the way we do it.

(On this regard, lately I have seen that tsx is easier to work with than ts-node. No need to change, just as a note for the future)

i18n/es.po
```diff
@@ -23,6 +23,32 @@ msgstr "cambio de unidades organizativas"
...
+msgid "No value"
+msgstr "Ninguno"
+
+msgid "Yes"
+msgstr ""
+
+msgid "No"
+msgstr ""
+
+#, fuzzy
```
tokland: [full file] review fuzzy tags and commented out translations

index.html
```diff
@@ -19,6 +19,10 @@
         <meta name="theme-color" content="#000000" />
 
         <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+        <!-- 
+            included here to avoid problems during build related to vite not able to support
+            commonjs libraries.
+          -->
```
tokland: what's included here?

src/data/entry-form/CustomForm.ts
```diff
@@ -0,0 +1,439 @@
...
+    } else if (dataElement && dataElement.categoryCombo) {
+        return dataElement.categoryCombo;
+    } else {
+        throw new Error(
+            `Cannot get category combo for dataSetElement: ${JSON.stringify(dataSetElement)}`
+        );
+    }
+}
+
+class Map {
```
tokland: `Map` shadows the built-in JS Map, we should probably use another name. 

Also, can we do it type-safe? (without any). Map<T> I guess, if the keys are always strings.

src/data/entry-form/CustomForm.ts
```diff
@@ -0,0 +1,439 @@
...
+    }
+}
+
+const map = (obj: any) => new Map(obj);
+
+const createViewDataElement = (de: DataElementViewTemplate) => ({
+    id: de.id,
+    displayFormName: de.displayName,
+    url: de.href,
+    hasUrl: () => !!de.href,
```
tokland: not important, but typically we use Boolean() instead of !!

src/data/entry-form/CustomForm.ts
```diff
@@ -0,0 +1,439 @@
...
+        helpers: {
+            getDataElementsByCategoryCombo,
+            getDataElementsByCategoryComboForIndicators,
+            createViewDataElement,
+            getHeaders,
+            getVisibleOptionCombos: getVisibleOptionCombos.bind(null, greyedFields),
+            getRowTotalId,
+        },
+        i18n: {
+            getString: (key: string) => {
```
tokland, line 246: let's move to separate function (and maybe use a Record instead of switch)

src/data/entry-form/custom-template.ts
```diff
@@ -0,0 +1,5 @@
...
...AgICAgICAgICAgICNpZiggJHNlY3Rpb24ubmFtZSApCiAgICAgICAgICAgICAgICAgICAgPGxpPjxhIGhyZWY9IiN0YWItJHtzZWN0aW9uLmlkfSI+JGVuY29kZXIuaHRtbEVuY29kZSggJHNlY3Rpb24ubmFtZSApPC9hPjwvbGk+CiAgICAgICAgICAgICAgICAjZWxzZQogICAgICAgICAgICAgICAgICAgIDxsaT48YSBocmVmPSIjdGFiLSR7c2VjdGlvbi5pZH0iPiRpMThuLmdldFN0cmluZygic2VjdGlvbiIpICR2ZWxvY2l0eUNvdW50PC9hPjwvbGk+CiAgICAgICAgICAgICAgICAjZW5kCiAgICAgICAgICAgICNlbmQKICAgICAgICA8L3VsPgogICAgICAgIAogICAgICAgICNmb3JlYWNoKCAkc2VjdGlvbiBpbiAkc2VjdGlvbnMgKQogICAgICAgICAgICA8ZGl2IGlkPSJ0YWItJHtzZWN0aW9uLmlkfSIgY2xhc3M9InR5cGUtJHtzZWN0aW9uLnR5cGV9Ij4KICAgICAgICAgICAgICAgIDxkaXYgY2xhc3M9ImluLXBlcmlvZCI+CiAgICAgICAgICAgICAgICAgICAgI3JlbmRlclNlY3Rpb24oICRzZWN0aW9uICkKICAgICAgICAgICAgICAgIDwvZGl2PgoKICAgICAgICAgICAgICAgIDxkaXYgY2xhc3M9Im91dC1vZi1wZXJpb2QiIHN0eWxlPSJkaXNwbGF5OiBub25lIj4KICAgICAgICAgICAgICAgICAgICAke2kxOG4uZ2V0U3RyaW5nKCAiY3VycmVudF9kYXRlX291dF9vZl9wZXJpb2QiICl9OgogICAgICAgICAgICAgICAgICAgIDxzcGFuIGNsYXNzPSJpbmZvIj48L3NwYW4+CiAgICAgICAgICAgICAgICA8L2Rpdj4KICAgICAgICAgICAgPC9kaXY+CiAgICAgICAgI2VuZAogICAgPC9kaXY+CiNlbHNlCiAgICAjZm9yZWFjaCggJHNlY3Rpb24gaW4gJHNlY3Rpb25zICkKICAgICAgICAjcmVuZGVyU2VjdGlvbiggJHNlY3Rpb24gJHNlY3Rpb24ubmFtZSApCiAgICAjZW5kCiNlbmQK";
```
tokland, line 2: What do you think of using Vite raw importer to keep these two files as .vm? I think it's something like this:

```
import customTemplateVelocity from "$/data/entry-form/custom-template.vm?raw";
```

src/data/repositories/CategoryComboD2Repository.ts
```diff
@@ -0,0 +1,44 @@
...
+            })),
+            optionsCombos: d2CategoryCombo.categoryOptionCombos.map(optionCombo => ({
+                id: optionCombo.id,
+                name: optionCombo.displayName,
+                options: optionCombo.categoryOptions.map(option => ({
+                    id: option.id,
+                    name: option.displayName,
+                })),
+            })),
+        });
```
tokland, line 42: Just as a note, the transformation displayName->name forces us to do a lot of traversals. d2-api exposes the DHIS2 API transformer "rename", which is very useful for this use case:

```
displayName: { $fn: { name: "rename", to: "name" } }
```

(which can be abstracted to avoid repetition, ofc)

src/data/repositories/DataSetD2Api.ts
```diff
@@ -208,7 +210,20 @@ export class DataSetD2Api {
         const projectDetails = projects.find(project => project.id === projectAttributeId);
         const dataElementGroups = this.buildDataElementsGroupsCodes(d2DataSet);
 
+        const disabledFields = d2DataSet.sections.flatMap(section => {
+            const degCode = this.extractCompetencyCode(section.id, section.code);
+            const coreCompetency = coreCompetencies.find(cc => cc.code === degCode);
+            return section.greyedFields.map((greyField): DataSet["disabledFields"][number] => {
+                return {
+                    competencyId: coreCompetency?.id ?? "",
```
tokland: Probably, that's not a problem here, but falling back to "" should always be considered carefully. There are different scenarios. In the scenario "this is something that should never happen," it's better to throw an exception (it's safer and keeps the code cleaner).

src/data/repositories/DataSetD2Repository.ts
```diff
@@ -140,44 +155,94 @@ export class DataSetD2Repository implements DataSetRepository {
...
+                const ids = categoryCombos.map(cc => cc.id);
+                return this.D2ApiCategoryCombo.getByIds(ids).map(categoryCombos => {
+                    const nonExistingCategoryCombos = differenceBy(
+                        d2CategoryCombos,
+                        categoryCombos,
+                        categoryCombo => categoryCombo.id
+                    );
+                    return {
+                        dataSetId: dataSet.id,
+                        categoryCombos: _(categoryCombos.concat(nonExistingCategoryCombos))
```
tokland: I was about to say "we can use _.concat as we already have a chain", but it does not exist :) Can you add it to Collection?

src/data/repositories/DataSetD2Repository.ts
```diff
@@ -326,128 +398,109 @@ export class DataSetD2Repository implements DataSetRepository {
...
-                    const dataElementsRelated = indicator.relatedDataElements.filter(
-                        dataElement => !dataElement.isComment
-                    );
+            return categoryCombosToCreate.map(categoryCombo => {
+                return {
+                    dataDimensionType: DIMENSITON_TYPE,
+                    publicAccess: "r-------",
+                    id: categoryCombo.id,
+                    name: categoryCombo.name,
+                    categories: categoryCombo.categories.map(category => ({ id: category.id })),
```
tokland: Not really important, but the future, this pattern is very typical, in many projects we have `getRefs(objs): Ref[]` in Base.ts

src/domain/entities/CategoryCombination.ts
```diff
@@ -2,30 +2,64 @@ import { Category } from "$/domain/entities/Category";
...
+        const categoriesIds = disaggregation?.categories.map(category => category.id) ?? [];
+        const categoriesIdsSets = new Set(categoriesIds);
 
-        const categoriesToFilter = dataElements.flatMap(dataElement => {
-            return dataElement.disaggregation?.categories.map(category => category.id) ?? [];
-        });
+        const allCategories = combinations
+            .filter(c => c.id !== disaggregation?.id)
+            .flatMap(combination => {
+                return combination.categories;
```
tokland: Sorry, it's very minor, but let's keep the arrow as a one-liner for consistency.

src/domain/entities/DataElement.ts
```diff
@@ -1,16 +1,17 @@
...
 export type DataElement = {
     id: Id;
     name: string;
     code: string;
-    disaggregation: Maybe<Disaggregation>;
+    description: string;
+    disaggregation: Maybe<DisaggregationAttrs>;
     categories: Category[];
     isComment: boolean;
+    valueType: string;
```
tokland, line 15: Typically, we want to narrow down these value types. However, if the domain does not use specific value types, we can consider it an "informative" field. Narrowing it down would require us to add validations and such, without providing much value. If that's the case, there's nothing to be done.

src/domain/entities/DataSetPeriodDate.ts
```diff
@@ -0,0 +1,10 @@
+import { PeriodDate } from "$/domain/entities/PeriodDate";
+import { Id } from "$/domain/entities/Ref";
+
+export type DataSetPeriodDate = {
+    id: Id;
+    name: string;
+    periodDate: PeriodDate;
+    outcomeDate: PeriodDate;
+    outputDate: PeriodDate;
```
tokland: At first glance, I am unsure about the naming. The suffix "date" suggests a plain date, but they are actually periods. EDIT: see also later comment on the class.

src/domain/entities/Indicator.ts
```diff
@@ -1,32 +1,107 @@
...
-import { COMMENT_SUFIX, DataElement, Disaggregation } from "$/domain/entities/DataElement";
+import { COMMENT_SUFIX, DataElement } from "$/domain/entities/DataElement";
 import { CoreCompetency } from "$/domain/entities/DataSet";
-import { Id } from "$/domain/entities/Ref";
+import { Id, NamedRef, Ref } from "$/domain/entities/Ref";
 import { HashMap } from "$/domain/entities/generic/HashMap";
 import { Struct } from "$/domain/entities/generic/Struct";
 import { Maybe } from "$/utils/ts-utils";
 
+export type DisaggregationAttrs = {
```
tokland: We typically use SomethingAttrs along with Struct. However, in this case, there is no struct; it's a plain object. I believe just using "Disaggregation" prevents confusion.

src/domain/entities/PeriodDate.ts
```diff
@@ -0,0 +1,52 @@
+import { Struct } from "$/domain/entities/generic/Struct";
+
+export type PeriodDetailsAttrs = { year: number; startDate: string; endDate: string };
+type PeriodDateAttrs = { startDate: string; endDate: string; periods: PeriodDetailsAttrs[] };
+
+export class PeriodDate extends Struct<PeriodDateAttrs>() {
```
tokland: In some other comment I had doubts about the naming. Maybe is the order? `DatePeriod` meaning "period of dates" seems more fitting.

src/domain/repositories/CategoryComboRepository.ts
```diff
@@ -0,0 +1,7 @@
+import { CategoryCombination } from "$/domain/entities/CategoryCombination";
+import { Id } from "$/domain/entities/Ref";
+import { FutureData } from "$/domain/entities/generic/Future";
+
+export interface CategoryComboRepository {
```
tokland: Sorry for the nit-picking: when doing text/file searches, it comes handy that entities/repos naming match: `CategoryCombinationRepository`

src/webapp/components/dataset-wizard/AddDisaggregateModal.tsx
```diff
@@ -104,5 +153,35 @@ export const AddDisaggregateModal = React.memo((props: AddDisaggregateModalProps
...
+        value: "all",
+        text: "All",
+    },
+    {
+        value: "individuals",
+        text: "All individuals in the data set",
+    },
+    {
+        value: "households",
+        text: "All HHS in the data set",
```
tokland: Do we need to i18n these texts?

src/webapp/components/dataset-wizard/DisaggregationStep.tsx
```diff
@@ -166,10 +196,29 @@ function useGetRelatedDataElements(props: { dataSet: DataSet; settings: DataSetS
                 const allIndicators = Indicator.buildAllIndicators(indicators);
                 setIndicatorsDataElements(allIndicators);
                 setRefresh(refresh + 1);
+                onChange(dataSet.setIndicators(indicators));
             }, console.error);
-    }, [compositionRoot.indicators.getRelated, dataSet, refresh, settings.existingIndicatorIds]);
+    }, [
+        compositionRoot.indicators.getRelated,
```
tokland: completely minor, but we typically just pass compositionRoot as dependency, as the whole object is immutable.

src/webapp/hooks/useDataSets.ts
```diff
@@ -90,6 +91,35 @@ export function useDeleteDataSets(props: DeleteDataSetsProps) {
...
+                        onSuccess();
+                    },
+                    err => {
+                        loading.hide();
+                        onError(err.message);
+                    }
+                );
+        },
+        [compositionRoot.dataSets.savePeriodDate, ids, onError, onSuccess, loading]
+    );
```
tokland: as discussed somewhere, this use-case is the one provided by useCallbackEffect.

tokland (commented):
This PR has some overlap with others I already reviewed, so I've tried not to repeat things I already said in them.

MiquelAdell:
@eperedo are all @tokland comments addressed? should we re-request a review? Thanks!

eperedo:
> @eperedo are all @tokland comments addressed? should we re-request a review? Thanks!

Not yet, working on it.

index.html
eperedo: Sorry, I removed the velocity.js script and forgot about the comment. We have a ESM version now.

src/data/entry-form/custom-template.ts
eperedo: I did this, but then I had some TS errors when running scripts: `Error: Cannot find module './sectionForm.vm?raw'` so loading then as base64 strings was my workaround. I tried to modify `scripts/tsconfig,json` to load these files but with no luck.

src/domain/entities/DataElement.ts
eperedo: Yes, we don't have any logic around this. Only used in the custom form, not really necessary to narrow the values yet.

package.json
```diff
@@ -112,7 +112,9 @@
         "localize": "yarn update-po && d2-i18n-generate -n d2-dataset-configuration -p ./i18n/ -o ./src/locales/",
         "update-po": "yarn extract-pot && find i18n/ -name '*.po' -exec msgmerge --backup=off -U {} i18n/en.pot \\;",
         "prepare": "husky install",
-        "migrate-ds-project": "npx ts-node src/scripts/migrate-dataset-project.ts"
+        "migrate-period-dates": "npx ts-node src/scripts/migrate-dataset-period-dates",
+        "migrate-ds-project": "npx ts-node src/scripts/migrate-dataset-project.ts",
+        "add-core-competency": "npx ts-node src/scripts/add-core-compentency-to-datasets.ts"
```
tokland, line 117: [nothing to do] This approach is OK for projects with few scripts (like this one). When the count starts to grow, it's more realistic to have a single "run-script": `yarn run-script src/scripts/migrate-dataset-project.ts ...`

src/data/D2ApiIndicator.ts
```diff
@@ -140,7 +204,7 @@ export class D2ApiIndicator {
                 );
 
                 return Indicator.create({
-                    measure: undefined,
+                    measure: "",
                     valueType: "",
```
tokland, line 208: No need to change as it adds little value at the present, but empty strings like the ones used for measure and valueType usually mean that the field should have been defined Maybe<string>.

src/data/entry-form/CustomForm.ts
```diff
@@ -296,19 +298,32 @@ const convertToSections = (
...
-    return result;
+    return result.sort((a, b) => {
+        const sectionA = getSectionName(a);
+        const sectionB = getSectionName(b);
+
+        const baseCompare = sectionA.localeCompare(sectionB);
+        if (baseCompare !== 0) return baseCompare;
+
+        return b.type.localeCompare(a.type);
+    });
```
tokland, line 324: [question] _.sortBy does not fulfill this use case because of the localCompare particularities?

src/data/entry-form/CustomForm.ts
```diff
@@ -324,42 +339,53 @@ function getItemsForSections(
...
+        .sortBy(indicator => indicator.displayName)
         .keyBy(indicator => indicator.id)
         .toObject();
 
     return items;
 }
 
 const getTemplate = (
-    dataset: any,
+    dataSet: any,
```
tokland: This is DataSetTemplate, right?

src/domain/repositories/AppSettingsDataRepository.ts
```diff
@@ -0,0 +1,6 @@
+import { AppSettingsData } from "$/domain/entities/AppSettingsData";
+import { FutureData } from "$/domain/entities/generic/Future";
+
+export interface AppSettingsDataRepository {
```
tokland, line 4: it's a bit confusing to have AppSettings and AppSettingsData, but I guess they hold conceptually different information. (Data -> usually too generic a name)

src/scripts/utils.ts
```diff
@@ -0,0 +1,6 @@
+export const escapeCSVField = (value: string): string => {
+    if (value.includes(",") || value.includes('"') || value.includes("\n")) {
+        return `"${value.replace(/"/g, '""')}"`;
+    }
+    return value;
+};
```
tokland: let's avoid generic "utils" file. This can go some file with name csv that only involves csv machinery.

src/webapp/components/dataset-wizard/DisaggregationStep.tsx
```diff
@@ -261,15 +234,16 @@ function useUpdateIndicators(props: {
...
+                const { indicator, dataElements } = indicatorDataElement;
                 const fullIds = joinIndicatorDataElementsId(indicatorDataElement);
                 if (!indicatorIdsByMode.includes(fullIds)) return indicatorDataElement;
 
                 const initialData =
                     indicator.type === "outputs"
-                        ? settings.indicators.filter(x => x.id === indicator.id)[0]?.disaggregation
-                        : dataElementsOutcomes.filter(de => de.id === dataElements[0]?.id)[0]
-                              ?.disaggregation;
+                        ? settings.indicators.filter(x => x.id === indicator.id)[0]
```
tokland: filter(...)[0] -> find

tokland (commented):
In-line comments.

(Not sure if it was me or the base branch, very big diff here)

src/data/entry-form/CustomForm.ts
eperedo: I tried but with no luck, because what I want here is to extract the section name then sort by section name and type, but for the type always output first than outcomes.

From this:

```ts
[
    {
        "name": "EDUCATION Outcomes",
        "type": "outcome"
    },
    {
        "name": "PROTECTION FROM VIOLENCE Outcomes",
        "type": "outcome"
    },
    {
        "name": "EDUCATION Outputs",
        "type": "output"
    },
    {
        "name": "PROTECTION FROM VIOLENCE Outputs",
        "type": "output"
    }
]
```

to this:

```ts
[
    {
        "name": "EDUCATION Outputs",
        "type": "output"
    },
    {
        "name": "EDUCATION Outcomes",
        "type": "outcome"
    },
    {
        "name": "PROTECTION FROM VIOLENCE Outputs",
        "type": "output"
    },
    {
        "name": "PROTECTION FROM VIOLENCE Outcomes",
        "type": "outcome"
    }
]
```

src/domain/repositories/AppSettingsDataRepository.ts
eperedo: This entity manages all dropdown components in the settings form. Initially I named it `AppSettingsOptions`, but it felt too generic so I renamed it to `AppSettingsData` 😅

src/data/entry-form/CustomForm.ts
tokland: I see, then I think _.orderBy is the abstraction: https://github.com/EyeSeeTea/dataset-configuration/blob/development-project-configurator/src/domain/entities/generic/__tests/Collection.spec.ts#L261

tokland (approved):
All good, code-wise.

PR #483, opened 2025-02-26
add configuration screen

eperedo:
### :pushpin: References

-   **Issue:** Closes 8697gqqjk https://app.clickup.com/t/8697gqqjk

### :memo: Implementation

We need to update the current settings in the dataStore since we have a new attribute for Period dates. Attached in clickup.

[Project-Configuration-app_config_app_settings.webm](https://github.com/user-attachments/assets/77a6601f-e4cd-41d8-a073-c32727562156)

#8697gqqjk


MiquelAdell (approved):

MiquelAdell (changes requested):
I get a test error:
Screenshot 2025-03-04 at 09 58 21
Does it need fixing?

eperedo:
> I get a test error: Screenshot 2025-03-04 at 09 58 21 Does it need fixing?

Fixed!

src/data/ApiSettingsCodec.ts
```diff
@@ -0,0 +1,38 @@
...
+        Codec.interface({
+            day: number,
+            month: number,
+        })
+    ),
+    periodLastYearEndDate: optional(
+        Codec.interface({
+            units: string,
+            value: number,
+        })
```
tokland: we can create a code `optionalPeriod` to keep it DRY

src/data/repositories/AppSettingsDataD2Repository.ts
```diff
@@ -0,0 +1,41 @@
...
+                indicatorGroups: { fields: metadataFields },
+                dataElementGroups: { fields: metadataFields },
+                dataElementGroupSets: { fields: metadataFields },
+                categoryCombos: { fields: metadataFields },
+                categories: { fields: metadataFields },
+                userGroups: { fields: metadataFields },
+            })
+        ).map(d2Response => {
+            return {
+                fields: d2Response.attributes.map(this.buildEntity),
```
tokland: This works because `buildEntity` does not use "this", but it will fail otherwise. We can create a local const so we can write: `buildEntities(d2Response.attributes)` (which calls this.buildEntity or whatever)

src/data/repositories/D2ApiAppSettings.ts
```diff
@@ -0,0 +1,248 @@
...
+                                    appSettings.periodDateField,
+                                    appSettings.indicatorHideField,
+                                    // used for scripts to migrate data
+                                    // not necessary to be configurable in the app through settings
+                                    metadataCodes.attributes.outcomeDates,
+                                    metadataCodes.attributes.outputDates,
+                                    metadataCodes.attributes.project,
+                                ])
+                                    .compact()
+                                    .value(),
```
tokland: this compact+value is repeated a lot and adds increases the line counts with no gain. Abstract the function to local const.

src/domain/entities/Config.ts
```diff
@@ -10,4 +10,8 @@ export type Config = {
     indicators: Indicator[];
     regions: Region[];
     userGroups: UserGroup[];
+    periodEndDateMonth: number;
+    periodEndDateDay: number;
+    periodLastYearEndDate: number;
+    periodLastYearUnits: string;
```
tokland: can we narrow down the units or is it impractical? I see we have a "UnitDate" (but in utils, if we finally use it, should be as entity)

src/utils/date.ts
```diff
@@ -37,3 +34,37 @@ export function addToDate(
...
+        value: String(i + 1),
+    }));
+}
+
+export function getUnits() {
+    return [
+        { text: "Day", value: "day" },
+        { text: "Month", value: "month" },
+        { text: "Year", value: "year" },
+    ];
```
tokland, line 55: if this is used in views, it should be translated, right?

src/utils/date.ts
```diff
@@ -37,3 +34,37 @@ export function addToDate(
...
+}
+
+export function getDaysPerMonthYear(
+    month: number,
+    year: number = new Date().getFullYear()
+): DropdownItem[] {
+    if (month === 0) return [];
+    const numDays = new Date(year, month, 0).getDate();
```
tokland, line 63: is that correct? Date month is 0-11, but the arg is 1-12, right?  (that's the problem with type number, we don't know). We all agree that Date using 0-11 was an ill-advised choice :)

src/webapp/components/modal-selector/ModalSelector.tsx
```diff
@@ -0,0 +1,104 @@
...
+        return (
+            <div style={{ ...style, width: "initial" }}>
+                <Grid item xs={12}>
+                    <Button onClick={() => onSelectItem(item)} color="primary" disableElevation>
+                        {item.text}
+                    </Button>
+                </Grid>
+            </div>
+        );
+    };
```
tokland: [best practices] In-line components are sub-optimal. Row should be a normal component. Now, you'll say, how do we access the variables in the closure? (like itemsToRender). It looks like <List /> has a prop itemData that is passed down to the children, so it should work.

src/webapp/components/modal-selector/ModalSelector.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+    const selectedValue = items.find(item => item.value === value);
+
+    return (
+        <>
+            <TextField
+                fullWidth
+                label={label}
+                onClick={() => setOpenModal(true)}
```
tokland: Not wrong, but in other projects we have useBooleanState, which encapsulates this typical use case in views.

src/webapp/components/settings-form/SettingsForm.tsx
```diff
@@ -0,0 +1,357 @@
...
+import { AppSettingsData } from "$/domain/entities/AppSettingsData";
+
+type SettingsFormProps = { onClose: () => void };
+
+export const SettingsForm = React.memo((props: SettingsFormProps) => {
+    const { onClose } = props;
+    const { compositionRoot, currentUser } = useAppContext();
+    const [settingsData, setSettingsData] = React.useState<AppSettingsData>();
+    const [appSettings, setAppSettings] = React.useState<AppSettings>();
+    const [selectedTab, setSelectedTab] = React.useState(0);
```
tokland: Can we used const strings for the tabs, instead of generic numbers? it will be more readable.

src/webapp/components/settings-form/SettingsForm.tsx
```diff
@@ -0,0 +1,357 @@
...
+            () => {
+                snackbar.success(i18n.t("Settings saved"));
+                onClose();
+                setLoading(false);
+            },
+            () => {
+                setLoading(false);
+            }
+        );
+    };
```
tokland: we have useCallbackEffect in other projects for callbacks which fire a side-effect.

src/webapp/components/settings-form/SettingsForm.tsx
```diff
@@ -0,0 +1,357 @@
...
+                    disabled={loading}
+                    color="primary"
+                    onClick={saveSettings}
+                    variant="contained"
+                    type="submit"
+                >
+                    {i18n.t("Save")}
+                </Button>
+            </DialogActions>
+        </Dialog>
```
tokland, line 134: add some blank lines to let the component breath

src/webapp/components/settings-form/SettingsForm.tsx
```diff
@@ -0,0 +1,357 @@
...
+                <Dropdown
+                    hideEmpty
+                    items={daysPerMonthYear}
+                    onChange={value => updateProject(Number(value), "periodEndDateDay")}
+                    label={i18n.t("Day")}
+                    value={String(appSettings.periodEndDateDay)}
+                />
+            </div>
+            <div style={{ display: "flex" }}>
+                <div style={{ flex: "1" }}>
```
tokland: move to some styles to avoid that this hardcoded styles re-render full components

tokland (commented):
In-line comments:

MiquelAdell:
@eperedo could you merge `development-project-configurator` into this branch to fix a conflict? 
thank you!

src/utils/date.ts
eperedo: This is correct because of the "0" value for the day. 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date

```
If the day of month is set to 0, it becomes the last day of the previous month.
```

>  We all agree that Date using 0-11 was an ill-advised choice :)

I blame java 😆 https://x.com/BrendanEich/status/771006397886533632

src/webapp/components/dataset-periods/DataSetPeriodDates.tsx
```diff
@@ -29,7 +30,7 @@ function useGetYears(props: { period: PeriodDate }) {
         return years.map((year): PeriodDate["periods"][number] => {
             const month = config.periodEndDateMonth;
             const day = config.periodEndDateDay;
-            const units = unitDates.find(ud => ud === config.periodLastYearUnits);
+            const units = getUnitDateFromString(config.periodLastYearUnits);
```
tokland: Now that periodLastYearUnits has a narrow type this is not required, right?

src/domain/entities/UnitDate.ts
```diff
@@ -0,0 +1,10 @@
+import { Maybe, UnionFromValues } from "$/utils/ts-utils";
+
+const unitDates = ["week", "month", "day"] as const;
+export type UnitDate = UnionFromValues<typeof unitDates>;
+
+export const DEFAULT_UNIT_DATE = "month" as const;
+
+export function getUnitDateFromString(unitValue: string): Maybe<UnitDate> {
+    return unitDates.find(ud => ud === unitValue);
+}
```
tokland: This is a helper not strictly relevant to the entity, I would move it to the data repository using it (if finally we keep it, see next comment)

src/data/ApiSettingsCodec.ts
```diff
@@ -1,5 +1,8 @@
 import { Codec, string, number, optional } from "purify-ts";
 
+const optionalPeriod = optional(Codec.interface({ day: number, month: number }));
+const optionalPeriodLastYear = optional(Codec.interface({ units: string, value: number }));
```
tokland: Would it make sense to narrow "units" also at this point? then we can completely remove the function that validates/parses it.

src/webapp/components/modal-selector/ModalSelector.tsx
```diff
@@ -29,20 +30,20 @@ const _ModalSelector = React.memo((props: ModalSelectorProps) => {
...
+    //     return (
+    //         <div style={{ ...style, width: "initial" }}>
+    //             <Grid item xs={12}>
+    //                 <Button onClick={() => onSelectItem(item)} color="primary" disableElevation>
+    //                     {item.text}
+    //                 </Button>
+    //             </Grid>
+    //         </div>
+    //     );
+    // };
```
tokland: Do we need to keep these comments? If yes, add a comment to the code explaining why.

src/webapp/components/modal-selector/ModalSelector.tsx
```diff
@@ -65,7 +66,7 @@ const _ModalSelector = React.memo((props: ModalSelectorProps) => {
             <ConfirmationDialog
                 open={openModal}
                 cancelText={i18n.t("Cancel")}
-                onCancel={() => setOpenModal(false)}
+                onCancel={() => modalActions.disable()}
```
tokland: One nice thing of using useBooleanState is that the actions can be directly used, and we get immutability for free: `onCancel={modalActions.disable}`

src/CompositionRoot.ts
```diff
@@ -89,6 +89,9 @@ function getCompositionRoot(repositories: Repositories, config: Config) {
             get: new GetAppSettingsUseCase(repositories.appSettingsRepository),
             save: new SaveAppSettingsUseCase(repositories.appSettingsRepository),
         },
+        dataElements: {
+            getByIds: new GetDataElementsByIdsUseCase(repositories.dataElementRepository),
+        },
```
tokland, line 94: I get diffs on i18n/en.pot, i18n/es.po after a yarn localize. Also, there are fuzzy strings.

tokland (commented):
In-line comments:

tokland (approved):

PR #485, opened 2025-03-08
send notification after create/update dataSet

eperedo:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869870ka7

### :memo: Implementation

After create/edit a dataSet:

- [x] Send a notification to global admins
- [x] Send a notification to user groups with suffix _M&EDatasetCompletion in name.

**Create dataSet Notification**

notification_message_create

**Edit dataSet Notification**
notification_message_edit

#869870ka7


MiquelAdell (approved):

i18n/es.po
```diff
@@ -387,6 +412,88 @@ msgstr ""
 msgid "Core Competencies"
 msgstr ""
 
+msgid "General"
+msgstr ""
+
+#, fuzzy
```
tokland: fuzzy strings

src/data/repositories/D2ApiMetadata.ts
```diff
@@ -151,6 +155,11 @@ export class D2ApiConfig {
                 periodEndDateMonth: appSettings.periodEndDateMonth,
                 periodLastYearEndDate: appSettings.periodLastYearEndDate,
                 periodLastYearUnits: appSettings.periodLastYearUnits,
+                userGroups: {
```
tokland, line 164: We can add `(settings): D2Config => {` up in the block to link the object to its type.

src/data/repositories/UserGroupD2Repository.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export class UserGroupD2Repository implements UserGroupRepository {
+    constructor(private api: D2Api) {}
+
+    getByNames(names: string[]): FutureData<UserGroup[]> {
+        return apiToFuture(
+            this.api.models.userGroups.get({
+                fields: { id: true, displayName: true, code: true },
+                filter: {
+                    name: { in: names },
```
tokland, line 15: not saying it's wrong, just as note: we are filtering by field "name" but we use the "displayName" in the get, if we have translations, will this be confusing for the user?

src/data/repositories/NotificationD2Repository.ts
```diff
@@ -0,0 +1,20 @@
...
+    constructor(private api: D2Api) {}
+
+    send(notification: NotificationMessage): FutureData<void> {
+        if (notification.recipients.length === 0) return Future.void();
+
+        return apiToFuture(
+            this.api.messageConversations.post({
+                subject: notification.title,
+                text: notification.body,
+                userGroups: notification.recipients.map(recipient => ({ id: recipient })),
```
tokland, line 16: Just as a note, the field is "recipients: Id[]", which is a bit generic, how do we know what kind of ids can we put there? Here in the implementation we see it will only work for userGroups. But maybe it's the way it is, we only need to notify userGroups and that's how we understand "recipients" in this implementation.

src/domain/usecases/SaveDataSetUseCase.ts
```diff
@@ -32,14 +48,121 @@ export class SaveDataSetUseCase {
...
+            .value();
+
+        const dataSetOrgUnits = _(dataSet.orgUnits)
+            .map(orgUnit => orgUnit.id)
+            .sort()
+            .value();
+
+        const areEqual = dataSetOrgUnits.every((value, index) => value === projectOrgUnits[index]);
+
+        return areEqual;
```
tokland: That's because we don't have _.isEqual, right? Maybe it's some declarative using _.difference + size ? (and then no need to sort)

src/domain/usecases/SaveDataSetUseCase.ts
```diff
@@ -32,14 +48,121 @@ export class SaveDataSetUseCase {
...
+    }
+
+    private sendNotificationError(
+        options: SaveDataSetOptions,
+        errorMessage: string
+    ): FutureData<void> {
+        const { dataSet, user } = options;
+
+        return this.getUsersGroups(options).flatMap(userGroups => {
+            const title = `There has been an error when dataset '${dataSet.name}' was being saved.`;
```
tokland: i18n?

src/webapp/components/dataset-wizard/SummaryDataSet.tsx
```diff
@@ -72,26 +73,46 @@ export const SummaryItem = React.memo((props: { label: string; value: string })
...
 });
 
+export const actions = ["edit", "create", "clone"] as const;
+export type DataSetRegisterAction = (typeof actions)[number];
+
+function getActionFromUrl(url: string): DataSetRegisterAction {
+    if (url.includes("edit")) return "edit";
+    if (url.includes("create")) return "create";
+    if (url.includes("clone")) return "clone";
+    throw new Error("Invalid action");
```
tokland: Typically we would not need this because we know where section are we, but only useSaveDataSet is using this, and no one using useSaveDataSet, so I cannot check what could be the workaround.

src/webapp/components/dataset-wizard/SummaryDataSet.tsx
```diff
@@ -72,26 +73,46 @@ export const SummaryItem = React.memo((props: { label: string; value: string })
...
+}
+
 export function useSaveDataSet(props: {
```
tokland: where is useSaveDataSet used?

tokland (commented):
In-line comments:

src/data/repositories/UserGroupD2Repository.ts
eperedo: This isn’t an issue for this particular use case since the group’s name isn’t used in the notification. However, it could be problematic if we need to display a list of user groups when filtering by name. I’m not entirely sure what the best solution is here:

- Update the implementation and use the name instead displayName
- Implement a more future-proof solution by including both name and translatedName in the entity

*Ideally this would be a `getByCode` method but in this server we don't have codes for userGroups.

src/data/repositories/NotificationD2Repository.ts
eperedo: Yes, for now this only works with userGroups. I've used this generic name in case we need to expand this in a future to other kind of recipients (like orgUnits or users)

tokland (approved):

MiquelAdell:
@eperedo I merged another PR and now this branch has some conflicts we'll need fixed before I can merge them to development-project-configurator could you please fix them? thanks!

eperedo:
> @eperedo I merged another PR and now this branch has some conflicts we'll need fixed before I can merge them to development-project-configurator could you please fix them? thanks!

Fixed!

PR #486, opened 2025-03-09
conditionally hide UI actions depending on user authorities

eperedo:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869870kpj

### :memo: Implementation

DataSets:

- Only users with these authorities can see the "Delete" action: `["F_DATASET_DELETE", "F_SECTION_DELETE", "F_INDICATOR_PUBLIC_ADD", "F_INDICATOR_PRIVATE_ADD"]`

- Only users with these authorities can see the "Create" action button: `["F_SECTION_DELETE", "F_SECTION_ADD", "F_DATASET_PRIVATE_ADD"]`

- Only users with these authorities can `Edit` a dataSet: `["F_SECTION_DELETE", "F_SECTION_ADD"]` and should have access to the current dataSet. These rules apply for other actions that involves editing a dataSet like: `Sharing Settings`, `Assign to Organisation Units`, `Set output/outcome period dates`

- Only admin users have access to the `Logs` action.

Category Combination:

- Only users with authority `["F_CATEGORY_COMBO_PUBLIC_ADD"]` can create category combinations in the `Disaggregation` step.

All of these rules are ignored for admin users.

#869870kpj


MiquelAdell (approved):

src/data/repositories/UserD2Repository.ts
```diff
@@ -8,22 +8,46 @@ export class UserD2Repository implements UserRepository {
...
+                    isAdmin ||
+                    authorities.dataSets.create.every(authority =>
+                        allAuthorities.includes(authority)
+                    ),
+                canEditCombinations:
+                    isAdmin ||
+                    authorities.categoryCombo.edit.every(authority =>
+                        allAuthorities.includes(authority)
+                    ),
+            },
```
tokland, line 37: the four are very similar, maybe create little local abstract function so keep it DRY and single-line?

src/data/repositories/UserD2Repository.ts
```diff
@@ -39,3 +63,18 @@ const userFields = {
...
+        createPublic: ["F_DATASET_PUBLIC_ADD"],
+        create: ["F_SECTION_DELETE", "F_SECTION_ADD", "F_DATASET_PRIVATE_ADD"],
+        delete: [
+            "F_DATASET_DELETE",
+            "F_SECTION_DELETE",
+            "F_INDICATOR_PUBLIC_ADD",
+            "F_INDICATOR_PRIVATE_ADD",
+        ],
+    },
+};
```
tokland, line 67: nice structure

src/domain/entities/DataSetList.ts
```diff
@@ -0,0 +1,24 @@
+import { DataSet } from "$/domain/entities/DataSet";
+import { User } from "$/domain/entities/User";
+import { Struct } from "$/domain/entities/generic/Struct";
+
+export type DataSetListAttrs = Pick<
+    DataSet,
+    "id" | "name" | "lastUpdated" | "permissions" | "canBeUpdated"
+>;
+
+export class DataSetList extends Struct<DataSetListAttrs>() {
```
tokland, line 10: Jorge does not recommend creating separate entities by complexity, but let's keep it for simplicity.

src/domain/entities/DataSetList.ts
```diff
@@ -0,0 +1,24 @@
...
+    get isPublic(): boolean {
+        return this.permissions.metadata.read;
+    }
+
+    get isPrivate(): boolean {
+        return !this.permissions.metadata.read;
```
tokland: minor, but if conceptually makes sense, it's nicer: !this.isPublic

src/domain/entities/Region.ts
```diff
@@ -13,5 +13,5 @@ export type Region = NamedCodeRef;
  */
 
 export function extractRegionCode(value: string): string {
-    return (value.slice(0, 2) || "").toUpperCase();
+    return (value?.slice(0, 2) || "").toUpperCase();
```
tokland: what is value? a code? let's call it code then.

src/domain/usecases/SaveDataSetUseCase.ts
```diff
@@ -14,52 +14,56 @@ import { UserGroup } from "$/domain/entities/UserGroup";
...
+                        .flatMap(() => {
+                            return this.saveProject(dataSet).flatMap(stats => {
+                                return this.sendNotification(options, stats.errorMessage);
+                            });
+                        })
+                        .flatMapError(error => {
+                            return this.sendNotificationError(options, error.message);
+                        });
+                });
+            });
```
tokland: minor, but unless variables are used in the chain, indentation can be reduced. For example:

```
           return this.validateDataSet(dataSet)
                .flatMap(() => {
                    return this.validateDataSetName(dataSet);
                })
                .flatMap(() => {
                    return this.dataSetRepository
                        .save([dataSet])
                        .flatMap(() => {
                            return this.saveProject(dataSet).flatMap(stats => {
                                return this.sendNotification(options, stats.errorMessage);
                            });
                        })
                        .flatMapError(error => {
                            return this.sendNotificationError(options, error.message);
                        });
                });
```

src/webapp/components/dataset-table/DataSetTableConfig.tsx
```diff
@@ -169,7 +185,9 @@ export function getCommonActions<T extends ReferenceObject>(
             text: i18n.t("Set output/outcome period dates"),
             icon: <DateRangeIcon />,
             multiple: true,
-            isActive: props.isActive,
+            isActive: records => {
+                return props.isActive(records) && canDataSetBeUpdated(records, user);
```
tokland: `props.isActive(records) && canDataSetBeUpdated(records, user)` repeated at least 4 times. Create an abstraction.

src/webapp/components/dataset-table/DataSetTableConfig.tsx
```diff
@@ -196,10 +214,18 @@ export function getCommonActions<T extends ReferenceObject>(
...
     ];
 }
+
+function canDataSetBeUpdated<T>(records: T[], user: User): boolean {
+    return records.every(dataSet => {
+        if (dataSet instanceof DataSetList) {
+            return dataSet.hasPermissionsToUpdate(user);
+        }
+    });
+}
```
tokland, line 228: This is a bit off. Instead of simply T or T extends Ref, we could add an extra restriction, saying it has to support hasPermissionsToUpdate(user). And then add the method for the other classes that can be passed here. This way, from the moment you call the function you know what things are expected from the records passed. And then you can remove the instanceof check. [see also next comment]

src/webapp/components/dataset-table/DataSetTableConfig.tsx
```diff
@@ -124,20 +133,23 @@ export type CommonActionsProps<T> = {
     isActive: (data: T[]) => boolean;
     onAction: (action: TableAction) => void;
     navigateTo: ReturnType<typeof useNavigateTo>;
+    user: User;
 };
 
 export function getCommonActions<T extends ReferenceObject>(
```
tokland: [related with the previous comment] This is going to be used by a restricted set of classes, right (A| B). Maybe can be explicit with that, instead of ReferenceObject.

tokland (commented):
In-line comments:

src/data/repositories/UserD2Repository.ts
```diff
@@ -21,32 +21,19 @@ export class UserD2Repository implements UserRepository {
...
-                    ),
-                canEditCombinations:
-                    isAdmin ||
-                    authorities.categoryCombo.edit.every(authority =>
-                        allAuthorities.includes(authority)
-                    ),
+                canCreatePublicDataSets: hasAccess(authorities.dataSets.createPublic),
+                canDeleteDataSets: hasAccess(authorities.dataSets.delete),
+                canCreateDataSets: hasAccess(authorities.dataSets.create),
+                canEditCombinations: hasAccess(authorities.categoryCombo.edit),
```
tokland, line 36: cool

src/webapp/components/dataset-table/DataSetTableConfig.tsx
```diff
@@ -136,20 +136,27 @@ export type CommonActionsProps<T> = {
     user: User;
 };
 
-export function getCommonActions<T extends ReferenceObject>(
+export interface HasPermission extends ReferenceObject {
+    hasPermissionsToUpdate(user: User): boolean;
+}
```
tokland, line 141: cool

src/webapp/components/dataset-table/DataSetTableConfig.tsx
```diff
@@ -136,20 +136,27 @@ export type CommonActionsProps<T> = {
...
+
+export function getCommonActions<T extends HasPermission>(
     props: CommonActionsProps<T>
 ): DataTableAction<T>[] {
     const { onAction, navigateTo, user } = props;
+
+    const isActionActive = (records: T[]): boolean => {
+        return props.isActive(records) && canDataSetBeUpdated(records, user);
+    };
```
tokland, line 150: No need to change, just as a note. Sometimes is more declarative to name variables not by how they will be used, but what they represent by themselves. For example: `canDataSetsBeUpdated` or whatever.

tokland (approved):

PR #490, opened 2025-04-08
replicate share settings from project to dataSet

eperedo:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/8697rj6fy

### :memo: Implementation

- [x] Get sharing settings for projects
- [x] Instead checking the first two letters of the project code now we replicate the same sharing settings from the project to the dataSet.

[Project-Configuration-app_multiple_sharing_settings.webm](https://github.com/user-attachments/assets/0d9021df-8ba8-43c8-89db-673f9547fe2a)

#8697rj6fy

### DIFF
--- a/src/data/D2ApiSharing.ts
+++ b/src/data/D2ApiSharing.ts
@@ -1,25 +1,13 @@
-import _ from "lodash";
 import { AccessData, AccessType } from "$/domain/entities/DataSet";
 import { Permission, Permissions } from "$/domain/entities/Permission";
+import { Id } from "$/domain/entities/Ref";
+import _ from "$/domain/entities/generic/Collection";
+import { HashMap } from "$/domain/entities/generic/HashMap";
 
 export class D2ApiSharing {
     buildPermission(permissions: string, permissionType: "data" | "metadata"): Permission {
-        switch (permissionType) {
-            case "metadata": {
-                const { canRead, canWrite } = this.buildPermissionByType(
-                    permissions,
-                    permissionType
-                );
-                return Permission.create({ read: canRead, write: canWrite });
-            }
-            case "data": {
-                const { canWrite, canRead } = this.buildPermissionByType(
-                    permissions,
-                    permissionType
-                );
-                return Permission.create({ read: canRead, write: canWrite });
-            }
-        }
+        const { canRead, canWrite } = this.buildPermissionByType(permissions, permissionType);
+        return Permission.create({ read: canRead, write: canWrite });
     }
 
     generateFullPermission(permissions: Permissions): OctalNotationPermission {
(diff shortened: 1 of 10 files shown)


src/data/D2ApiSharing.ts
```diff
@@ -0,0 +1,132 @@
+import _ from "lodash";
```
tokland: We use using est-lodash for this project, right?

src/data/D2ApiSharing.ts
```diff
@@ -0,0 +1,132 @@
...
+                return Permission.create({ read: canRead, write: canWrite });
+            }
+            case "data": {
+                const { canWrite, canRead } = this.buildPermissionByType(
+                    permissions,
+                    permissionType
+                );
+                return Permission.create({ read: canRead, write: canWrite });
+            }
+        }
```
tokland: Aren't the two branches identical?

src/data/D2ApiSharing.ts
```diff
@@ -0,0 +1,132 @@
...
+
+export type D2ApiSharingFields = {
+    owner: string;
+    external: boolean;
+    users: D2AccessRecords;
+    userGroups: D2AccessRecords;
+    public: string;
+};
+
+type D2AccessRecords = Record<string, D2ApiSharingName>;
```
tokland: let's create a type alias for the key (string), so we know what it represents.

src/data/repositories/DataSetD2Repository.ts
```diff
@@ -314,10 +317,15 @@ export class DataSetD2Repository implements DataSetRepository {
             }
 
             const existingAttributes = existingDataSet?.attributeValues;
+            const sharingData = this.d2ApiSharing.generateSharingData(dataSet);
 
             const result = {
```
tokland: naming: result -> can we change it to a more descriptive name?

type:  anonymous object, let's put an explicit type.

src/domain/entities/Permission.ts
```diff
@@ -20,6 +20,13 @@ export class Permission extends Struct<PermissionAttrs>() {
                 : Permission.create({ read: true, write: false }),
         };
     }
+
+    static initialPermissions(): Permissions {
```
tokland: more declarative: noPermissions ?

(it's not used anywhere, but we can keep it)

src/domain/entities/Project.ts
```diff
@@ -12,9 +13,19 @@ export type ProjectAttrs = {
     dataSets: DataSet[];
     lastUpdated: ISODateString;
     orgsUnits: OrgUnit[];
+    access: AccessData[];
 };
 
 export class Project extends Struct<ProjectAttrs>() {
+    get uniqueAccessCodes() {
```
tokland: public interface, let's put an explicit return type.

src/domain/entities/Project.ts
```diff
@@ -12,9 +13,19 @@ export type ProjectAttrs = {
...
+        return _(this.access)
+            .filter(access => access.type === "groups")
+            .map(access => Project.extractCode(access.name))
+            .filter(accessCode => accessCode.length > 0)
```
tokland: Now that I see this filter over the accessCode, check if it makes sense for extractCode to return "string | undefined". As we know, returning empty strings to signal "no value" is typically a red flag.

tokland (commented):
In-line:

tokland (approved):

PR #492, opened 2025-05-05
add MLF filters

eperedo:
### :pushpin: References

- **Issue:** Closes https://app.clickup.com/t/8698wjj9t

### :memo: Implementation

Metadata and details about how to configure groupSet and groups in the clickup task.

[Project-Configuration-app_MLF_filters.webm](https://github.com/user-attachments/assets/ee5da471-d832-4722-979f-16e82c6fa8f9)

#8698wjj9t

### DIFF
--- a/i18n/es.po
+++ b/i18n/es.po
@@ -1,7 +1,7 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: i18next-conv\n"
-"POT-Creation-Date: 2025-04-02T18:02:53.605Z\n"
+"POT-Creation-Date: 2025-05-07T15:49:30.462Z\n"
 "PO-Revision-Date: 2018-10-25T09:02:35.143Z\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
@@ -88,9 +88,6 @@ msgstr ""
 
 msgid "DataSet not found"
 msgstr "DataSet no encontrado"
-
-msgid "Data set name already exists: {{dataSetName}}"
-msgstr "Ya existe un dataset con el nombre: {{dataSetName}}"
 
 msgid "There has been an error when dataset '{{dataSetName}}' was being saved."
 msgstr ""
@@ -300,6 +297,18 @@ msgid "Origin"
 msgstr ""
 
 msgid "Type"
+msgstr ""
+
+msgid "Error loading MLF's: {{error}}"
+msgstr ""
+
+msgid "Loading..."
+msgstr "Cargando..."
+
+msgid "Master Log Frames"
+msgstr ""
+
+msgid "No MLF's available for this dataSet"
 msgstr ""
 
 msgid "Additional"
@@ -547,6 +556,3 @@ msgstr ""
 
 msgid "Updating period date"
 msgstr ""
-
-msgid "Loading..."
-msgstr "Cargando..."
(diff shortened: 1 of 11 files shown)


i18n/es.po
```diff
@@ -302,6 +299,18 @@ msgstr ""
...
+msgstr ""
+
+msgid "Loading..."
+msgstr "Cargando..."
+
+msgid "Master Log Frames"
+msgstr ""
+
+msgid "No MLF's available for this dataSet"
+msgstr ""
```
tokland, line 312: Not sure what's the policy with these translations: is the app being used in Spanish? Do we enforce adding the translations of the languages we know?

src/data/repositories/MasterLogFrameD2Repository.ts
```diff
@@ -0,0 +1,169 @@
...
+
+    private isDataElementGroup(
+        group: D2DataElementGroup | D2IndicatorGroup
+    ): group is D2DataElementGroup {
+        return "dataElements" in group;
+    }
+
+    private isIndicatorGroup(
+        group: D2DataElementGroup | D2IndicatorGroup
+    ): group is D2IndicatorGroup {
```
tokland: As a rule of thumb, we try to avoid type guards in code under our control. I guess the necessity for this is having `d2Group: D2DataElementGroup | D2IndicatorGroup`. A typical approach is adding a field "type" to these two types, and now we have 100% type-safe discrimination.

src/data/repositories/MasterLogFrameD2Repository.ts
```diff
@@ -0,0 +1,169 @@
...
+    description: true,
+};
+
+type D2DataElementGroup = MetadataPick<{
+    dataElementGroups: { fields: typeof dataElementGroupFields };
+}>["dataElementGroups"][number];
+
+type D2IndicatorGroup = MetadataPick<{
+    indicatorGroups: { fields: typeof indicatorGroupFields };
+}>["indicatorGroups"][number];
```
tokland: [note] As it's normal, we use this pattern throughout the code-base, but it's not terribly nice to read, so we can create an abstraction (types/d2-api.ts) that captures this typical use case without all the noise. In this case, it would look: 

`type D2IndicatorGroup = D2RecordType<"indicatorGroups", typeof indicatorGroupFields>` 

(`D2RecordType` is a proposal, it's a hard to find a name).

src/domain/entities/MasterLogFrame.ts
```diff
@@ -0,0 +1,9 @@
+import { IndicatorType } from "$/domain/entities/Indicator";
+import { Id, Ref } from "$/domain/entities/Ref";
+
+export type MasterLogFrame = {
```
tokland, line 14: What's a Master Log Frame? Maybe we can add some TSDocs here, explaining its role in the app.

src/domain/usecases/GetMasterLogFrameUseCase.ts
```diff
@@ -0,0 +1,11 @@
+import { MasterLogFrame } from "$/domain/entities/MasterLogFrame";
+import { FutureData } from "$/domain/entities/generic/Future";
+import { MasterLogFrameRepository } from "$/domain/repositories/MasterLogFrameRepository";
+
+export class GetMasterLogFrameUseCase {
+    constructor(private masterLogFrameRepository: MasterLogFrameRepository) {}
+
+    execute(codes: string[]): FutureData<MasterLogFrame[]> {
```
tokland: When use case is called (GetMasterLogFrameUseCase), it won't be very clear what are the strings argument being passed. I'd use an `options` here.

tokland (commented):
In-line:

i18n/es.po
eperedo: My bad. I forgot to ask you about this @MiquelAdell. Do we have any translation for MLF's?

tokland (approved):

PR #500, opened 2025-06-28
Yearly companion rule

gqcorneby:
### :pushpin: References

-   **Issue:** Closes #86999261b [Make mandatory companions "year dependent"](https://app.clickup.com/t/86999261b)

### :memo: Implementation
- Add companion rule yearly parsing ("scope") 
- Updated project model to indicate start and end date
- added yearly filter
- updated summary companion indicator warning

### NOTES
- New notation for yearly rule:
    - `<year> => <rule> ; <year> => <rule>`
- still works for 1 rule. Will treat that as scope "default". 

https://github.com/user-attachments/assets/92c0fde6-8dcf-4d79-a647-b7f4436cfff0

image
Added 2024 and 2025 to:
[GL-E913/E: Total value of cash/voucher distributed for education in first-line response purposes (USD)](https://gorsdev.nrc.no/dhis-web-maintenance/index.html#/edit/dataElementSection/dataElement/N4iE2JXU2aE)
- same rule
[GL-F801/AS: # of individuals trained in skills related to Food and Agricultural Systems ](https://gorsdev.nrc.no/dhis-web-maintenance/index.html#/edit/dataElementSection/dataElement/jCWTSQO5DZs)
- removed one AND rule for 2025


src/data/D2IndicatorCompanionRuleParser.ts
```diff
@@ -0,0 +1,40 @@
...
+import { IndicatorCompanionRule } from "$/domain/entities/Indicator";
+import { D2CompanionRuleParser } from "$/data/D2CompanionRuleParser";
+
+export class D2IndicatorCompanionRuleParser {
+    constructor(private readonly ruleText: string) {}
+
+    public parse(): Maybe<IndicatorCompanionRule> {
+        if (!this.ruleText?.trim()) return undefined;
+
+        return this.ruleText.includes("=>")
```
eperedo: I know is very unlikely we decide to change this symbol "=>", but can we extract into a variable so we can modify it in one place.

src/data/D2IndicatorCompanionRuleParser.ts
```diff
@@ -0,0 +1,40 @@
...
+            .split(";")
+            .map(s => s.trim())
+            .filter(Boolean);
+
+        const parsedEntries = _(entries).compactMap(entry => {
+            const [scopePart, rulePart] = entry.split("=>").map(s => s.trim());
+            if (!scopePart || !rulePart) return null;
+
+            const parsed = new D2CompanionRuleParser(rulePart).buildCompanionRule();
+            return parsed ? [scopePart, parsed] as const : null;
```
eperedo: Can we return "undefined" instead null? Just to be consistent with the rest of the code.

src/domain/entities/Indicator.ts
```diff
@@ -275,38 +280,95 @@ export class Indicator extends Struct<IndicatorAttrs>() {
...
-    getCompanionCodesByType(type: "output" | "outcome"): Code[] {
+    getCompanionCodesScope(): HashMap<Code, IndicatorCompanionScope[]> {
+        const outcomeCodes = this.getCompanionCodesByType("outcome");
+        const outputCodes = this.getCompanionCodesByType("output");
+        return _([...outcomeCodes.keys(), ...outputCodes.keys()])
+            .uniq()
+            .flatMap(scope => {
+                return _([...(outcomeCodes.get(scope) ?? []), ...(outputCodes.get(scope) ?? [])])
+                    .uniq()
+                    .map(code => [code, scope] as [Code, IndicatorCompanionScope]);
```
eperedo: We tried to avoid explicit cast, Are we doing this to avoid the undefined error in the groupFromMap?

If that's the reason I think we can throw an error because that means we have some invalid data, right? Maybe I'm missing something.

```ts
.groupFromMap(([code, scope]) => {
    if (!code || !scope) throw new Error("Invalid code or scope");
    return [code, scope];
});
```

src/domain/entities/Indicator.ts
```diff
@@ -275,38 +280,95 @@ export class Indicator extends Struct<IndicatorAttrs>() {
...
-    buildCompanionRuleMessage(): { outcomeMessage: string; outputMessage: string } {
-        const outcomeMessage = buildCompanionRuleMessage(this.companionRules?.outcomeRule);
-        const outputMessage = buildCompanionRuleMessage(this.companionRules?.outputRule);
-        return { outcomeMessage, outputMessage };
+    validateCompanionRules(indicatorByCodes: HashMap<LowercaseString, Indicator>): {
+        outcomeRulesValid: HashMap<IndicatorCompanionScope, boolean>;
+        outputRulesValid: HashMap<IndicatorCompanionScope, boolean>;
+    } {
+        const evaluateIndicatorRule = (rule: CompanionRule) => evaluateRule(rule, indicatorByCodes);
+        const outcomeRuleIsValid: HashMap<IndicatorCompanionScope, boolean> = this.companionRules
```
eperedo: Normally for these cases we don't need to specify the type because it should infer from the method `processCompanionRuleByScope`.

I know here could be tricky because the function will complain about `HashMap.empty()` not being `HashMap<IndicatorCompanionScope, boolean>`.

Two proposals here:

- return `undefined` and then validate in the `SummaryDataSet` component. (longer and more code, but I think more easy to understand [subjective])

```ts
// you'll need to modify the return type in the function validateCompanionRules
// outcomeRulesValid: Maybe<HashMap<IndicatorCompanionScope, boolean>>
// a custom type could be useful here

const outputRuleIsValid = this.companionRules?.outputRule
  ? processCompanionRuleByScope(this.companionRules.outputRule, evaluateIndicatorRule)
  : undefined;
            
  // usage
  const { outcomeRulesValid } = indicator.validateCompanionRules(indicatorByCodes);
  const outcomeValidations = outcomeRulesValid  ? outcomeRulesValid.toPairs(...) : []
```

- set the type in the `HashMap.empty` method. (shorter, but kinda cryptic because by reading the return type I don't really expect an empty object)

```ts
const outcomeRuleIsValid = this.companionRules?.outcomeRule
  ? processCompanionRuleByScope(this.companionRules.outcomeRule, evaluateIndicatorRule)
  : HashMap.empty<IndicatorCompanionScope, boolean>();
```

What do you think? Feel free to challenge this since this is a subjective thing.

src/utils/ts-utils.ts
```diff
@@ -62,6 +62,13 @@ export function recordOf<T>() {
     };
 }
 
+export type LowercaseString = string & { __lowercase: true };
```
eperedo, line 65: Do we need this `__lowercase`? And the explicit case in the `toLowercaseString` function?

src/webapp/components/dataset-wizard/FilterCompanionIndicators.tsx
```diff
@@ -0,0 +1,28 @@
...
+export type FilterCompanionIndicatorsProps = {
+    onFilterChange: (value: ChipItem[]) => void;
+    scopes: ChipItem[];
+    scopeValue: string[];
+    hidden: boolean;
+};
+
+export const FilterCompanionIndicators = React.memo((props: FilterCompanionIndicatorsProps) => {
+    const { onFilterChange, scopes, scopeValue, hidden: hide } = props;
+    return (
```
eperedo, line 18: we can use an early return here:

```tsx
export const FilterCompanionIndicators = React.memo((props: FilterCompanionIndicatorsProps) => {
    const { onFilterChange, scopes, scopeValue, hidden: hide } = props;

    if (hide) return null;

    return (
        <ChipFilter
            items={scopes}
            label={i18n.t("Showing suggested indicators for years")}
            onChange={onFilterChange}
            value={scopeValue}
            mode={"multiple"}
        />
    );
});
```

src/webapp/components/dataset-wizard/SummaryDataSet.tsx
```diff
@@ -64,47 +66,78 @@ export const SummaryList = React.memo((props: { dataSet: DataSet }) => {
...
+        const scopes = indicator.keys();
+
+        const scopeText = React.useCallback(
+            (scope: string) => {
+                return scope === requiredScope && scope !== "default" ? (
+                    <strong>
+                        {scope}:{i18n.t("The companion rules for this year need to be satisfied")}
+                    </strong>
+                ) : (
+                    scope
```
eperedo: Not a problem, but can we put a tag for the scope? 

`<p>scope</p>`. 

`span` also works if you need an inline element.

src/webapp/components/dataset-wizard/SummaryDataSet.tsx
```diff
@@ -121,43 +154,91 @@ const useGetMissingSelectedCompanion = (props: { dataSet: DataSet }) => {
...
+                            ),
+                        })
+                    ),
+                    ...outputRulesValid.toPairs().map(
+                        ([scope, isValid]): ValidationItem => ({
+                            code: indicator.code,
+                            scope,
+                            type: "outputs" as const,
+                            isValid,
+                            message: buildValidationMessages(
```
eperedo: Similar logic for creating a `ValidationItem`. Can we extract into a function? buildValidationItemByType(type: "outcome" | "outputs")

eperedo (changes requested):
[code review only]
Looks good, thanks @gqcorneby. Minor things about TS. Check my comments and let me know if you have any questions.

src/domain/entities/Indicator.ts
gqcorneby: I think it is safe to remove. Sorry, I was trying different things at the time and I may have forgotten to clean this up. Thank you!

src/utils/ts-utils.ts
gqcorneby: For this one, I was encountering a bug during testing and eventually traced it to the validation and the passed `indicatorByCodes` (`validateCompanionRules(indicatorByCodes: HashMap<LowercaseString, Indicator>)`. 
Some of the codes had mismatching cases which caused false invalid states. I tried introducing this lowercase typed string using branding so case insensitivity is more explicit.

src/domain/entities/Indicator.ts
gqcorneby: Hi Eduardo! For me, I would opt to have a consistent return type and use HashMap.empty().

Since we'd perform null checks for values, I feel like there's no need to overload the absence of rules with undefined. If we ever want to short-circuit behavior based on an empty result, we can add a HashMap isEmpty() or similar.

eperedo (approved):
All good. Thanks @gqcorneby !

MiquelAdell (approved):

PR #505, opened 2025-09-03
Fix: add period dates and update data entry

gqcorneby:
### :pushpin: References

-   **Issue:** Closes #869a9hcp5 [Issue with the dates on the new app](https://app.clickup.com/t/869a9hcp5)

### :memo: Implementation
- generate and save dataInputPeriod
    - `dataInputPeriod` is generated based on attribute `GL_INTERVAL_DATES (Data Input Dates)` 
    - Opening and closing dates are the min startDate and max endDate of `GL_DATASET_PERIOD_DATES (Period Dates)`
- recalculate and save `openFuturePeriods` when selecting project or updating data input dates
- update custom form to use `GL_DATASET_PERIOD_DATES`
- Update date handling:
    - use `toISODateWithoutTimezone` when saving
    - With the current flow, saved short dates are -1 day because of timezone
        1. convert attribute short date (+8 timezone) -> `new Date(2025, 3, 1, 0, 0, 0) April 1 2025 00:00:00 GMT+8`
        2. in saving, we convert this date `toISOString()` then get the date value, which will be `March 31 2025`
            -  `new Date(2025, 3, 1, 0, 0, 0).toISOString()` is `2025-03-31T16:00:00.000Z`

TODO: 
- [ ] confirm period data value when project updates
    - should default be current periodDate or undefined (if project doesn't have start/end dates)

**CREATE DATA SET**

https://github.com/user-attachments/assets/d6e9f592-ea70-4c30-9cfc-8f87aea718f4

**UPDATE 2027 CLOSING (fields disable fields)**

https://github.com/user-attachments/assets/fb8598aa-4d55-4f6a-be1c-d5dabb8df93e

https://github.com/user-attachments/assets/d3dc77a2-90e2-409d-9f3d-1ba80e1ff4bb

### DIFF
--- a/src/domain/entities/DatePeriod.ts
+++ b/src/domain/entities/DatePeriod.ts
@@ -1,6 +1,6 @@
 import { Struct } from "$/domain/entities/generic/Struct";
 import { Config } from "$/domain/entities/Config";
-import { addToDate, formatDateToISO, getDiff, stringToTime } from "$/utils/date";
+import { addToDate, toISODateWithoutTimezone, getDiff, stringToTime } from "$/utils/date";
 import _, { Collection } from "$/domain/entities/generic/Collection";
 import { Maybe } from "$/utils/ts-utils";
 
@@ -71,8 +71,8 @@ export class DatePeriod extends Struct<DatePeriodAttrs>() {
                 return MonthlyPeriodDetails.create({
                     year: targetYear,
                     month: targetMonth,
-                    startDate: formatDateToISO(openingDate),
-                    endDate: formatDateToISO(closingDate),
+                    startDate: toISODateWithoutTimezone(openingDate),
+                    endDate: toISODateWithoutTimezone(closingDate),
                 });
             })
             .value();
@@ -113,7 +113,7 @@ export class DatePeriod extends Struct<DatePeriodAttrs>() {
     private buildShortFormat(date: string) {
         if (!date) return "";
 
-        const datePart = formatDateToISO(new Date(date)).split("T")[0];
+        const datePart = toISODateWithoutTimezone(new Date(date)).split("T")[0];
         if (!datePart) return "";
         return datePart.replace(/-/g, "");
     }
--- a/src/domain/entities/DatePeriod.ts
+++ b/src/domain/entities/DatePeriod.ts
@@ -41,14 +41,10 @@ export class DatePeriod extends Struct<DatePeriodAttrs>() {
             return [];
         }
 
-        const allStartTimes = _(periods)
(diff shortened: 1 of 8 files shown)


src/domain/entities/DatePeriod.ts
```diff
@@ -24,10 +35,85 @@ export class DatePeriod extends Struct<DatePeriodAttrs>() {
         }));
     }
 
+    get dataInputPeriods(): MonthlyPeriodDetails[] {
+        const { periods, startDate: startDateStr, endDate: endDateStr } = this;
+        if (!periods.length) {
+            return [];
+        }
+
+        const allStartTimes = _(periods)
```
eperedo: I think we can extract this functionality into a private method and reuse it for start/end Times.

```ts
_(periods)
  .compactMap(p => stringToTime(p.startDate))
  .value();
```

Also I think the name of the variable should be `allEndTimeS` to be consistent with `allStartTimes`, right?

src/utils/date.ts
```diff
@@ -37,6 +38,37 @@ export function addToDate(date: string, units: UnitDate, unitValue: number): str
...
+            if (dateA !== dateB) {
+                const dateADay = dateA.getDate();
+                const dateBDay = dateB.getDate();
+                const daysInCurrentMonth = new Date(
+                    dateA.getFullYear(),
+                    dateA.getMonth() + 1,
+                    0
+                ).getDate();
+                const dayFraction = (dateBDay - dateADay) / daysInCurrentMonth;
+                monthDiff += dayFraction;
```
eperedo: Not a big deal, but normally we try to avoid mutating values. We can directly call to `return monthDiff + dayFraction` 

Also I think the condition should be `if (dateA.getTime() !== dateB.getTime())` to compare the actual dates and not the reference, right?

Example:

```ts
var a = new Date("2024-12-05");
var b = new Date("2024-12-05");
console.log(a !== b) // true
console.log(a.getTime() !== b.getTime()) // false
```

eperedo (changes requested):
[code review]. Looks good, just two minor observations. Thanks @gqcorneby

src/domain/entities/DatePeriod.ts
gqcorneby: Thanks, Eduardo! Minor thing but I went back and forth with this :laughing: I thought it might be overkill to extract it as a method. Will apply :+1:

src/utils/date.ts
gqcorneby: Good catch! Thank you!

eperedo (approved):

MiquelAdell (approved):

PR #509, opened 2025-10-13
Feature/secton config

gqcorneby:
### :pushpin: References

-   **Issue:** Closes #869atk4wy [Row and column configuration is missing](https://app.clickup.com/t/869atk4wy)

### :memo: Implementation
- add section config in table actions menu

https://github.com/user-attachments/assets/20a214d9-e375-4ed9-bfe8-3b2de219de92

- Already has beta version but noting here that requires this update in d2-ui-component: https://github.com/EyeSeeTea/d2-ui-components/pull/252


MiquelAdell (approved):

src/webapp/components/dataset-wizard/IndicatorsDataSet.tsx
```diff
@@ -618,3 +684,14 @@ const ToggleButtonStyled = styled(ToggleButton)`
 const SuggestCompanionContainer = styled.div`
     padding: 0.5em !important;
 `;
+
+const SectionConfigContainer = styled(Box)`
+    padding: 0.5em 1.5em 0 1.5em;
```
eperedo: can we use logical properties here? https://github.com/EyeSeeTea/know-how/wiki/UI-Guide#using-css-logical-properties

src/webapp/components/dataset-wizard/IndicatorsDataSet.tsx
```diff
@@ -618,3 +684,14 @@ const ToggleButtonStyled = styled(ToggleButton)`
...
+    display: flex;
+    flex-direction: column;
+    min-width: 270px;
+`;
+
+const CheckboxSectionConfig = styled(Checkbox)`
+    margin: 8px 0 !important;
```
eperedo: Same for margin

src/webapp/components/dataset-wizard/IndicatorsDataSet.tsx
```diff
@@ -290,6 +298,9 @@ export const IndicatorsDataSet = React.memo((props: IndicatorsDataSetProps) => {
                         selection={dataSet.indicators.map(indicator => ({ id: indicator.id }))}
                         onChange={validateIndicators}
                         sorting={sorting}
+                        customConfig={
+                            dataSet && <SectionConfig dataSet={dataSet} onChange={onChange} />
```
eperedo: Isn't this check always truthy since `dataSet` is already fetched in the parent component?

eperedo (changes requested):

src/webapp/components/dataset-wizard/IndicatorsDataSet.tsx
gqcorneby: You're right! I was initially working with a Maybe and didn't notice to update this. Thanks!

gqcorneby:
Just noting that the error in the workflow is because I pushed the usage of the updated prop in d2-ui-components. Once we have a beta release of that, the workflow should be fine.

eperedo (approved):
As @gqcorneby said we need the latest ui-component version to avoid the typescript error, but code-wise looks good to me so approving. Thanks!

PR #517, opened 2025-12-10
Feature: disable "not affected by displacement" 2026

gqcorneby:
### :pushpin: References

-   **Issue:** Closes #869bcnh4t [disable category option "not affected by displacement"](https://app.clickup.com/t/869bcnh4t)

### :memo: Implementation
- disable checkboxes that involve the category option `bvFA7fsiN3T` and if project startdate is 2026 

https://github.com/user-attachments/assets/b4258a61-f86e-48a5-a8b0-2fbf3398d6b5


MiquelAdell (approved):

src/webapp/components/dataset-wizard/grey-fields/components.tsx
```diff
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { CSSProperties } from "react";
```
eperedo: Not a big deal, but we normally only import "React" and then use the properties directly `React.CSSProperties`.

src/webapp/components/dataset-wizard/grey-fields/components.tsx
```diff
@@ -476,3 +511,10 @@ export const getKey = (categoryCombo: Ref, categoryOptions: Category["options"])
         .value();
     return [categoryCombo.id, ...sortedUniqueIds].join(".");
 };
+
+export function disableCategoryOptionFor2026(optionIds: string[], projectStartDate?: string) {
+    if (!projectStartDate) return false;
+
+    const startDate = new Date(projectStartDate);
+    return startDate.getFullYear() >= 2026 && optionIds.includes("bvFA7fsiN3T");
```
eperedo: Maybe you could include the link to the issue? So we can understand the decision behind using a hardcoded reference here.

eperedo (approved):
A couple of minor comments, none of them affect functionality so I'm approving.

gqcorneby:
Hi @MiquelAdell! I've added the fix for the following reported issues
1. grey fields not disabled when editing
2. disabled fields still visible in data entry app


https://github.com/user-attachments/assets/fd318e92-d9c7-49e3-b474-b4201e0cdc56


https://github.com/user-attachments/assets/23f7eeec-ff7a-4792-87fe-3b36615906d8

gqcorneby:
Hi @eperedo and @MiquelAdell! I'm requesting review again because there's been some changes that I think aren't minor. Here are the issues being addressed: 
1. grey fields not disabled when editing - `dataSet.project.startDate` was empty
2. disabled fields still visible in data entry app - need to set `dataSet.disabledField`

I had to update the approach to add disabled fields upon selection of the indicators so the disabling isn't dependent the using allowing step 4 to load. Note: user can jump from one step to another after first pass.


https://github.com/user-attachments/assets/0085dd3e-43c5-4181-bb23-faa64694d7ff

PR #518, opened 2026-02-18
add script to move old projects to project_old category

eperedo:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869c225cj #869c225cj

### :memo: Implementation

Migrate Projects to Old Projects:

**Step 1 (typescript script):**

- [x] Import sqlviews metadata: `metadata_old_projects_in_cc_deprecated_sqlview.json`
- [x] Script to move old projects from category **Project (MRwzyV0kXv9)** to **Old_Project (WIWj6TauYO8)**

```shell
npm run remove-old-projects -- --dhis2-url="http://localhost:8080" --username="admin" --password="district" --persist
```

**Step 2 (typescript script):**

- [x] Add categories **Old_Project, CC_Deprecated and TargetActual** to categoryCombo **Project-CC-Target/Actual_Old**
- [x] Update categoryCombo for affected categoryOptionCombos from **Project-CC-Target/Actual -> Project-CC-Target/Actual_Old**. Datavalues are not modified.

```shell
npm run move-deprecated-cocs -- --dhis2-url="http://localhost:8080" --username="admin" --password="district" --persist
```

After this we must **clear cache** from DHIS2:

image

**Before Migration**

You can see a lot of projects in the selector

image

image

**After migration**

Only projects with no dataValues in deprecated core competencies

image

Creating a new dataSet pointing to the old category combo **Project-CC-Target/Actual_Old** and you can see the same dataValues previoulsy assigned to the combo **Project-CC-Target/Actual**

image

If we got an error in the Data Entry app we'll need to check and delete possible duplicated cocs in the Project-CC-Target/Actual_Old categoryCombo. This can be done with the sql script attached in the issue **fix_duplicate_category_option_combos_v1.sql**


src/scripts/sql/metadata_old_projects_in_cc_deprecated_sqlview.json
```diff
@@ -0,0 +1 @@
...yoptionid = core_co.categoryoptionid\nJOIN dataelementcategory core_cat\n  ON core_cat.categoryid = cat_core.categoryid\n AND core_cat.name = 'Phase of Emergency'\nJOIN categoryoptioncombos_categoryoptions cocco\n  ON cocco.categoryoptionid = core_co.categoryoptionid\nJOIN categorycombos_optioncombos ccoc\n  ON ccoc.categoryoptioncomboid = cocco.categoryoptioncomboid\nJOIN categorycombo cc\n  ON cc.categorycomboid = ccoc.categorycomboid\n AND cc.uid = 'GmXXE8fiCK5'\nJOIN datavalue dv\n  ON dv.attributeoptioncomboid = cocco.categoryoptioncomboid\n AND dv.value IS NOT NULL\n AND dv.value != ''\n AND dv.deleted = false\nJOIN categoryoptioncombos_categoryoptions cocco_proj\n  ON cocco_proj.categoryoptioncomboid = cocco.categoryoptioncomboid\nJOIN dataelementcategoryoption project_co\n  ON project_co.categoryoptionid = cocco_proj.categoryoptionid\nJOIN categories_categoryoptions cat_proj\n  ON cat_proj.categoryoptionid = project_co.categoryoptionid\nJOIN dataelementcategory proj_cat\n  ON proj_cat.categoryid = cat_proj.categoryid\n AND proj_cat.name = 'Project'\nWHERE core_co.name LIKE '%[DEPRECATED]%'","type":"QUERY","cacheStrategy":"NO_CACHE","id":"XKCUtfKjjOt","attributeValues":[]}]}
```
tokland, line 1: I'd add the payload formatted. Additionally, to keep nice diffs, we can process it like this: `jq --sort-keys`

tokland, line 1: [nothing to do] Ideally, what we would have is the SQL file and a script that would generate (or post) the JSON payload.

EDIT: ok, I see that we actually have that as a SQL file below. I guess we have the JSON so it's easier to distribute.

package.json
```diff
@@ -114,7 +114,8 @@
         "prepare": "husky install",
         "migrate-period-dates": "npx ts-node src/scripts/migrate-dataset-period-dates",
         "migrate-ds-project": "npx ts-node src/scripts/migrate-dataset-project.ts",
-        "add-core-competency": "npx ts-node src/scripts/add-core-compentency-to-datasets.ts"
+        "add-core-competency": "npx ts-node src/scripts/add-core-compentency-to-datasets.ts",
+        "remove-old-projects": "npx tsx src/scripts/move-old-projects.ts"
```
tokland: tsx should be added as devDependency, right? and I guess no need to have both, so we could remove ts-node and use tsx in all scripts.

src/scripts/move-old-projects.ts
```diff
@@ -0,0 +1,206 @@
...
+    password: string;
+    persist: boolean;
+}>;
+
+type Category = Readonly<{ id: Id; categoryOptions?: Array<{ id: Id }> }>;
+
+const DEFAULTS = {
+    sqlViewId: "XKCUtfKjjOt",
+    fromCategoryId: "MRwzyV0kXv9",
+    toCategoryId: "WIWj6TauYO8",
```
tokland: let's add a comment with the names for these entries so we see in a glance what's the default action.

`fromCategoryId: "MRwzyV0kXv9" // Some category`

src/scripts/move-old-projects.ts
```diff
@@ -0,0 +1,206 @@
...
+
+    console.debug(`Metadata result: ${metadataSummary}`);
+};
+
+const getOldProjectsWithDeprecatedData = async (
+    api: D2Api,
+    sqlViewId: Id
+): Promise<{ id: string; name: string }[]> => {
+    // @ts-expect-error
+    const response = await api.sqlViews.query(sqlViewId, {}, { paging: false }).getData();
```
tokland: `@ts-expect-error` is typically used for type testing, as a way of saying "there MUST be an error on the next line, otherwise something is wrong". Here I assume you want `@ts-ignore`.

However, if you tagged the line because of the non-support prop `paging`, it's safer to pass it as a string so we still type-check the rest of the code. This should work: `{ ["paging" as string]: false }`

src/scripts/move-old-projects.ts
```diff
@@ -0,0 +1,206 @@
...
+type ScriptArgs = Readonly<{
+    url: string;
+    username: string;
+    password: string;
+    persist: boolean;
+}>;
+
+type Category = Readonly<{ id: Id; categoryOptions?: Array<{ id: Id }> }>;
+
+const DEFAULTS = {
```
tokland: `DEFAULTS` looks weird, not only because typically a "default" can be overridden, but we should find a non-generic name that describes what it is instead. `metadata` or similar?

src/scripts/move-old-projects.ts
```diff
@@ -0,0 +1,206 @@
...
+    return _(category.categoryOptions ?? [])
+        .map(categoryOption => categoryOption.id)
+        .uniq()
+        .value();
+};
+
+const buildUpdatedCategoryWithoutIds = (category: Category, idsToRemove: Array<Id>): Category => {
+    const currentIds = getCategoryOptionIds(category);
+    const nextIds = _(currentIds)
+        .filter(id => !idsToRemove.includes(id))
```
tokland, line 158: no need to change, just as a general note, typically _.union, _.intersection, _.difference operations are more declarative than filter+include checks.

src/scripts/sql/sqlview_projects_with_dv_in_cc_deprecated.sql
```diff
@@ -0,0 +1,30 @@
+SELECT DISTINCT project_co.uid   AS project_uid,
```
tokland, line 1: filename: sqlview.sql. We should be more specific with the name, the operation performed by this sql.

src/scripts/sql/sqlview_projects_with_dv_in_cc_deprecated.sql
```diff
@@ -0,0 +1,30 @@
...
+JOIN dataelementcategory core_cat
+  ON core_cat.categoryid = cat_core.categoryid
+ AND core_cat.name = 'Phase of Emergency'
+JOIN categoryoptioncombos_categoryoptions cocco
+  ON cocco.categoryoptionid = core_co.categoryoptionid
+JOIN categorycombos_optioncombos ccoc
+  ON ccoc.categoryoptioncomboid = cocco.categoryoptioncomboid
+JOIN categorycombo cc
+  ON cc.categorycomboid = ccoc.categorycomboid
+ AND cc.uid = 'GmXXE8fiCK5'
```
tokland, line 16: comment with what this uid stands for

src/scripts/sql/project_cc_target_actual_migration.sql
```diff
@@ -0,0 +1,263 @@
...
+  ON cocco.categoryoptioncomboid = dv.attributeoptioncomboid
+JOIN dataelementcategoryoption opt
+  ON opt.categoryoptionid = cocco.categoryoptionid
+ AND opt.name LIKE '%[DEPRECATED]%'
+WHERE dv.deleted = false;
+
+-- COMMIT;
+
+-- DROP TABLE IF EXISTS old_projects;
+-- DROP TABLE IF EXISTS cocs_to_move;
```
tokland: Were these two drops also meant to be commented out?

tokland (commented):
[code-only review, as requested by PM]

The workflow looks good to me. About code details:

src/scripts/sql/project_cc_target_actual_migration.sql
eperedo: Deleted. It was necessary when the sql take care of moving the categories as well, but now we're doing it with the TS script from the step 1.

src/scripts/sql/metadata_old_projects_in_cc_deprecated_sqlview.json
eperedo: Yes, because eventually we'll need to import this on dev and prod.

tokland (approved):

PR #522, opened 2026-07-17
fix(dataset-wizard): disable 2026 displacement option for outcomes

gqcorneby:
### :pushpin: References

-   **Task:** #869bcnh4t [disable category option "not affected by displacement"](https://app.clickup.com/t/869bcnh4t)

### :memo: Implementation
**Issue**: The 2026 hardcode that disables the "Not affected by displacement" category option (`bvFA7fsiN3T`) worked for output indicators but not outcomes. Outcome sections render the indicator's **related data elements** (different ids), so a disabled field keyed on the outcome *indicator* id never matched the rendered cells and so the column kept showing in the data-entry form.

-   `getDisableTargetsWith2026bvFA7fsiN3T` builds disabled-field targets keyed on the **rendered** element: `indicator.id` for outputs, each non-comment `relatedDataElement.id` for outcomes.
-   Outcome `relatedDataElements` are resolved (via `indicators.getRelated`) at indicator selection. Gated to 2026 projects and only when needed so the disabling applies even if the disaggregation step is skipped.
-   The feature stays isolated in `useDisable2026bvFA7fsiN3T.ts` for clean removal after April 2026.

https://github.com/user-attachments/assets/fc839439-e05a-49cd-86b9-578c10c52f24

- Select project "AFFM2605 (DANIDA)"
- Select outcome indicator `GL-EB101/Ab`.
- Save the dataset, then open the DHIS2 **Data Entry** form: the "Not affected by displacement (Not to be used in 2026)" column should be gone from **both** the EDUCATION Outputs and EDUCATION Outcomes tabs (previously it still showed on Outcomes).
-   The other displacement statuses (IDP, Refugee, Host Community, Returnee, Other, Conflict-affected but not displaced) must remain.

### DIFF
--- a/src/webapp/components/dataset-wizard/useDisable2026bvFA7fsiN3T.ts
+++ b/src/webapp/components/dataset-wizard/useDisable2026bvFA7fsiN3T.ts
@@ -1,8 +1,9 @@
 import _ from "$/domain/entities/generic/Collection";
 import React from "react";
-import { Indicator } from "$/domain/entities/Indicator";
+import { Indicator, IndicatorType } from "$/domain/entities/Indicator";
 import { DataSet, DisabledField } from "$/domain/entities/DataSet";
 import { Id } from "@eyeseetea/d2-api";
+import { Maybe } from "$/utils/ts-utils";
 import { HashMap } from "$/domain/entities/generic/HashMap";
 import { CategoryCombination } from "$/domain/entities/CategoryCombination";
 import { useAppContext } from "$/webapp/contexts/app-context";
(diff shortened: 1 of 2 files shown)


src/webapp/components/dataset-wizard/useDisable2026bvFA7fsiN3T.ts
```diff
@@ -100,30 +42,119 @@ export function useDisable2026bvFA7fsiN3TOnIndicatorUpdate() {
...
+                    indicator.type === "outcomes" &&
+                    indicator.relatedDataElements.length === 0 &&
+                    !existingIndicatorIds.includes(indicator.id)
+            );
+            if (!needsResolution) return Future.success(indicators);
+
+            return compositionRoot.indicators.getRelated.execute({
+                dataSet,
+                indicatorIdsToIgnore: existingIndicatorIds,
+            });
```
eperedo: Since we are introducing an async method now, should we at least log the error to the console here? Just in case of a network error we can detect the source faster.

```ts
.mapError(error => {
    console.error("Could not resolve relatedDataElements for outcomes", error.message);
});
```

eperedo (approved):
Looks good! Just a minor comment but non merge blocking.

Thanks @gqcorneby !

